PR #16, opened 2019-04-30
Do not compress the contents of the tile

jpalermo:
Since releases are already compressed, a lot of time is spent
compressing/decompressing that does not significantly change the
file size.

### DIFF
--- a/builder/zipper.go
+++ b/builder/zipper.go
@@ -29,7 +29,7 @@ func (z Zipper) Add(path string, file io.Reader) error {
 
 	return z.add(&zip.FileHeader{
 		Name:   path,
-		Method: zip.Deflate,
+		Method: zip.Store,
 		Modified: time.Now(),
 	}, file)
 }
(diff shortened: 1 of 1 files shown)


mcwumbly:
I'm curious to see some numbers for both *time* and *file size* before/after this change.

For example:

- PAS build tile time (before/after)
- PAS file size (before/after)
- PAS upload tile time (before/after)

jpalermo:
SF-PAS 2.5.2

 _ | Compress | No Compress
-----|------------ | -------------
Build Time | 280 Seconds | 52 Seconds
File Size | 10868436704 bytes | 10865551694 bytes
Total Upload Processing Time | 338 Seconds | 289 Seconds
Upload time after Rails recieved request | 150 Seconds | 108 Seconds

It's faster to build, faster to upload, and somehow even slightly a smaller file.

You can actually just delete the lines entirely, `zip.Store` is the default. I thought it might be good to leave the `zip.Store` to show intention to future code archeologists.

rizwanreza:
Tests passed. Merging.

mcwumbly:
> It's faster to build, faster to upload, and somehow even slightly a smaller file.

Thank you for capturing that data here. It too may help future archaeologists.

PR #17, opened 2019-05-03
Remove regex from assets.yml

rowanjacobs:
In lieu of using a configurable regex in `assets.yml`, this PR hard-codes the regex.

The new hard-coded regex is `^.+/(?P<release_name>[a-z-_0-9]+)-(?P<release_version>v?[0-9\.]+(-\w+)??)-(?P<stemcell_os>([a-z_]*-?){1,2})-(?P<stemcell_version>\d+\.\d+)(\.0)?\.tgz$`. In particular, this matches `-lts` release versions, and is agnostic as to S3 directory structure.

There are future improvements that can be made to configurability, some of which might involve removing the regex now that it's an internal implementation detail.

We're making this change because seriously, look at that regex, it's awful, are we really asking users to configure and understand it? 🙀

### DIFF
--- a/commands/fetch_test.go
+++ b/commands/fetch_test.go
@@ -25,7 +25,6 @@ compiled_releases:
   region: us-west-1
   access_key_id: mykey
   secret_access_key: mysecret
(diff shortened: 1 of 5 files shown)


mcwumbly:
lgtm (from product perspective)

the concerns I had earlier look like they've been addressed.

(glad we're just removing something from the public API but not adding or changing anything)

PR #19, opened 2019-05-09
Allow stemcell criteria to be specified in metadata [fixes #18]

### LINKED ISSUE #18: `kiln bake` fails if `--stemcell-tarball` is not provided.
This was a regression in 0.19.0. Previous versions allowed the stemcell criteria to be manually specified in the `metadata.yml` file, but 0.19.0 and later versions fail with the following message if neither `--assets-file` or `--stemcell-tarball` are specified:

```
Reading stemcell criteria from .lock
2019/05/02 15:31:33 could not execute "bake": failed to parse stemcell: open .lock: no such file or directory
```

pianohacker:
As of 0.18.x, one of --assets-file or --stemcell-tarball is required; without either, `kiln bake` fails with `open .lock: no such file or directory`.

This change allows the previous behavior, where the user manually specifies the stemcell criteria in their metadata.yml.


mcwumbly:
thanks for the PR @pianohacker -- one question I have from skimming it... do we specify what happens when stemcell criteria is not provided anywhere? if not, what do you think the behavior should be in that case?

cc @rizwanreza @syslxg

pianohacker:
> thanks for the PR @pianohacker -- one question I have from skimming it... do we specify what happens when stemcell criteria is not provided anywhere? if not, what do you think the behavior should be in that case?
> 
> cc @rizwanreza @syslxg

I don't know if it's formally specified. The previous behavior (now codified in an acceptance test) was to leave any `stemcell_criteria` in the metadata unchanged. I think this makes sense; if the user isn't using the `$( stemcell )` helper, they likely want to specify the criteria manually.

robertjsullivan:
@syslxg  Could you please review this?  We'd like to get it merged in.

mcwumbly:
@robertjsullivan thanks for the ping again. I put this on our board to discuss at standup tomorrow to see who can review this.

rizwanreza:
Thanks for the PR @pianohacker.

PR #24, opened 2019-06-13
Adds support for --stemcells-directory flag in kiln

### LINKED ISSUE #23: Add multiple --stemcell-tarball to bake command
We are using kiln bake command to build product for OpsMan like this:
```
kiln bake \
    ...
    --stemcell-tarball "$( find "p-pks-integrations/stemcells" -maxdepth 1 -name 'light-bosh-stemcell-*-google-kvm-ubuntu-*-go_agent.tgz'  -print -quit )" \
    ...
```
However if we need to support more than one stemcell (for windows and ubuntu) and kiln doesn't allow this.

Could the kiln command support a `--stemcell-directory` option as an argument to supply multiple stemcells? Here is an example: 

```
kiln bake \
    ...
    --stemcell-directory "p-pks-integrations/stemcells" \
    ...
```

spinfooser:
- Backwards compatible with past usage of stemcell
   helper: "$( stemcell )"
- Change stemcell template helper to take osname as
   an argument for multiple stemcells only
- Deprecate stemcell-tarball in favor of stemcell-directory
- Resolves #23

Co-authored-by: Josh Mattila <jmattila@pivotal.io>
Co-authored-by: Maryam Labib <mlabib@pivotal.io>


rizwanreza:
1. please restore the acceptance test for `--stemcell-tarbal`
We noticed that we removed `--stemcell-tarball` tests here, but we think we should keep at least one end-to-end test for that case until we decide to remove it. 

2. please allow parameter to the `stemcell` template helper to be optional when there's one stemcell:
It also seems unclear that choosing to use `--stemcells-directory` is what makes it compulsory to use `$( stemcell <os-name>)` format. It would be nice to keep the old functionality there if there's one stemcell only. This will allow better backwards compatibility for this new feature. Even if we decide to keep this new behavior, let's add the acceptance test back for `--stemcell-tarball` so it's clear that the `$( stemcell )` usage in the template requires that there be no arguments.  To clarify: regardless of how `kiln`  is being invoked, if there's one stemcell, we should not require the template author to specify which os to select (`$( stemcell )`).

3. when both the `--stemcell-tarball` and `--stemcells-directory` are specified, please fail-fast explaining to the user that these flags are mutually exclusive.



-- @pivotaljohn @garrettcheadle

spinfooser:
@pivotaljohn @garrettcheadle This feedback makes sense. 

As for #1, I was wondering if I should do this. I'll add one back to make sure it keeps working in future commits.

#2 shouldn't be hard to add. 

#3 I didn't even think about adding this, I'll add a test at the top level for this.

Thanks for the feedback.

spinfooser:
@pivotaljohn @garrettcheadle Hey, we just pushed the changes you requested to the branch, and updated the git commit message. If you could review it, that would be great!

pivotaljohn:
@spinfooser @maryamklabib — your fresh bits are avail as an official release: https://github.com/pivotal-cf/kiln/releases/tag/0.21.0

Thanks for your contribution!

PR #27, opened 2019-07-06
add documentation for fetch command and assets.* files

crhntr:
Please review this before merging. I am not sure about all the wording and there may still be grammar/spelling problems. I will review myself on Monday with fresh eyes.

- I tend to like to add some PAS RelEng context where it adds value. Since this is a general tool, how do y'all feel about that?
- I added some examples. They worked for me. Please check if they work for you. Should I use real name of the LastPass note we use for PAS example in the story? It would make it easier for PAS contributors to discover our process.


README.md
```diff
@@ -10,7 +10,118 @@ both in CI environments and in command-line to produce a tile.
...
+The following are **required** in the `compiled_releases` hash in an assets
+file.
+
+- `type`: must be "s3"
+- `bucket`: must be the name of the s3 bucket
+- `region`: must be the region of the bucket
+- `access_key_id`: must be an IAM access key id that has read permission for the
+  specified bucket
+- `secret_access_key`: must be the secret for the specified `access_key_id`
+- `regex:`: is used to specify a capture groups to map release and stemcell
```
pivotaljohn: Suggestion: "a regular expression applied to the full-path directory listing of the S3 bucket.  Only files that match the regex will considered.  This regular expression must include four named capture groups..."

README.md
```diff
@@ -10,7 +10,118 @@ both in CI environments and in command-line to produce a tile.
...
+  - `release_name` must map to the assets.lock file under `releases[].name`
+  - `release_version` must map to the assets.lock file under
+    `releases[].version`
+  - `stemcell_os` must map to the assets.lock file under `stemcell_criteria.os`
+  - `stemcell_version` must map to the assets.lock file under
+    `stemcell_criteria.version`
+
+### Assets Lock
+
+Currently the releases for assets lock file is can not be generated by kiln.
```
pivotaljohn: As a reader, I'm expecting to first get a definition of what this file is and does.  For example: "This file contains the full list of specific versions of all releases that will go into the tile AND the target stemcell."

README.md
```diff
@@ -10,7 +10,118 @@ both in CI environments and in command-line to produce a tile.
...
+compiled_releases:
+ type: s3
+ bucket: compiled-releases
+ region: us-west-1
+ access_key_id: $(variable "aws_access_key_id")
+ secret_access_key: $(variable "aws_secret_access_key")
+ regex: ^2.6/.+/(?P<release_name>[a-z-_0-9]+)-(?P<release_version>v?[0-9\.]+)-(?P<stemcell_os>[a-z-_]+)-(?P<stemcell_version>\d+\.\d+)(\.0)?\.tgz$
+```
+
+*Storing credentials like S3 keys in a repo is probably not the best idea, so
```
pivotaljohn: I think we can use even stronger language: "Credentials like S3 keys are not stored in git repos.  To support separating that information from non-sensitive configuration, you can reference variables like you do in tile config."

README.md
```diff
@@ -10,7 +10,118 @@ both in CI environments and in command-line to produce a tile.
...
+a local directory specified by the `--releases-directory` flag. It discovers
+releases based on information from both the assets file and an assets lock file.
+The assets lock file name is expected to be a file in the same directory as the
+specified assets file with `lock` instead of `yaml` as the filename extension.
+The S3 object name is determined based on using regular expression capture
+groups.
+
+Kiln verifies that the checksum (SHA1) of the downloaded release matches
+checksum specified for the release in the assets lock file. If the checksums do
+not match, then the releases that don't match will be deleted from disk. *Since
```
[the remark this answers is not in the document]
crhntr: I am not sure about this behavior. I read the code a little and it seems like it will only delete the release if the downloaded release does not match the checksum. If a dev build is in the releases directory it will be ignored.

README.md
crhntr: I like how this reads better. Thanks!

README.md
crhntr: This makes sense. I am not sure if adding some clarification about what target stemcell means would also be good. I will copy in your text. If you would like to commit to master with a little more about what that means, that might be nice for some new users.

crhntr:
I believe this can be merged. A quick read through would be appreciated. :)

rizwanreza:
LGTM :+1:

PR #31, opened 2019-08-02
add reasable default and ability to make releases dir

crhntr:
`--releases-directory` default is "releases"
`--assets-file default` default is "assets.yml"


evanfarrar:
LGTM

PR #38, opened 2019-11-12
add retry loop and increase timeout for ListUserGroups HTTP Request

crhntr:
Authored-by: Christopher Hunter <chunter@pivotal.io>

[#169495176] kiln publish fails with Network Timeout error


crhntr:
This is a quick fix. We should revisit this when we do this story: https://www.pivotaltracker.com/story/show/169626270

Issue #36, opened 2019-11-07
`go get github.com/pivotal-cf/kiln` fails with a compilation error


benchristel:
I'm not sure why this happens, because `go build` inside the `kiln` repo on fec14b69 is able to produce a `kiln` executable. Still, we get this error on macOS and linux:

```
$ go get github.com/pivotal-cf/kiln
# github.com/pivotal-cf/kiln/commands
../kiln/commands/publish.go:123:27: cannot use client.Releases (type *pivnet.ReleasesService) as type PivnetReleasesService in assignment:
	*pivnet.ReleasesService does not implement PivnetReleasesService (wrong type for List method)
		have List(string, ...pivnet.QueryParameter) ([]pivnet.Release, error)
		want List(string) ([]pivnet.Release, error)
```

gregttn:
We have the same issue.

benchristel:
FWIW, I heard from a coworker yesterday that she had a very similar problem with a totally different repo. So this might not be specific to Kiln.

coro:
I believe this commit is the issue: https://github.com/pivotal-cf/go-pivnet/commit/19ab81bdc63c7b1f5eb3701d3e06e617d1bcf5db

It looks like the public method `ReleasesService.List()` changed its interface in this commit in a breaking fashion, without reflecting this in an appropriate semver major version bump. This means that the interface implemented in this library is no longer implemented by the type `ReleasesService`.

I'm raising an issue on that repo now.

rszumlakowski:
Hi.  I'm on the Pivnet team.  We'll look into this issue.

coro:
@benchristel looks like this is now fixed, I can successfully:

```bash
GO111MODULE=on go get github.com/pivotal-cf/kiln
```

crhntr:
This issue looks like it was fixed a while ago. I will close it now.

PR #39, opened 2019-11-20
Bump go-pivnet to v2.0.11

gregttn:
This addresses issue https://github.com/pivotal-cf/kiln/issues/36

Co-authored-by: Connor Rogers <crogers@pivotal.io>


gregttn:
Worth noting: we ran all tests locally with `ginkgo -r .` and they all passed.

Issue #43, opened 2020-04-22
`kiln bake --stub-releases` can easily create tiles that do not upload properly


mcwumbly:
When using `kiln bake --stub-releases` with most the active `p-runtime` branches, they tile they create cannot be uploaded to ops manager without an error like this:

```
$ om upload-product -p srt-2.9.2-build.1.pivotal
processing product
beginning product upload to Ops Manager
 581.16 KiB / 581.16 KiB [==========================================] 100.00% 0s

could not execute "upload-product": failed to upload product: request failed: unexpected response:
HTTP/1.1 422 Unprocessable Entity
...
```

Somewhere in the lengthy error output, you'll see some output like this:

```
replace /tmp/ops_manager/d20200422-735-1gpnlnh/releases? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL\n(EOF or read error, treating as \"[N]one\" ...)
```

Just attempting to unzip the file, you get similar prompts a few times:
```
 extracting: releases/cf-networking-2.22.0-ubuntu-xenial-315.13.tgz
replace releases? [y]es, [n]o, [A]ll, [N]one, [r]ename:
...
 extracting: releases/metric-registrar-1.0.4-ubuntu-xenial-315.13.tgz
replace releases? [y]es, [n]o, [A]ll, [N]one, [r]ename:
...
 extracting: releases/syslog-11.4.0-ubuntu-xenial-315.13.tgz
replace releases? [y]es, [n]o, [A]ll, [N]one, [r]ename:
```

Looking more closely at the output from `kiln bake`, I notice a few lines that say `Adding releases to /tmp/...` with no subdirectory:

```
Adding releases/cf-networking-2.22.0-ubuntu-xenial-315.13.tgz to /tmp/srt.Y2yWW4PaC/srt-2.9.2-build.1.pivotal...
Adding releases to /tmp/srt.Y2yWW4PaC/srt-2.9.2-build.1.pivotal...
Adding releases/cflinuxfs3-0.82.0-ubuntu-xenial-315.13.tgz to /tmp/srt.Y2yWW4PaC/srt-2.9.2-build.1.pivotal...
Adding releases/metric-registrar-1.0.4-ubuntu-xenial-315.13.tgz to /tmp/srt.Y2yWW4PaC/srt-2.9.2-build.1.pivotal...
Adding releases to /tmp/srt.Y2yWW4PaC/srt-2.9.2-build.1.pivotal...
[...]

PR #44, opened 2020-04-23
Create stub metadata when no tgz file exists

mcwumbly:
Prior to this change, tiles created with --stub-releases would often
fail to upload properly to Ops Manager.

This is problematic since the --stub-releases
feature exists primarily to aid tile developers in iterating more
quickly on UI changes with a real Ops Manager.

https://github.com/pivotal-cf/kiln/issues/43
https://www.pivotaltracker.com/story/show/172467082


matt-royal:
Thank you for the pull request, @mcwumbly. This is a small change and has test coverage, so I'll merge it and then confirm that the tests pass CI.

PR #242, opened 2021-03-04
kiln can pre-process tile repos

crhntr:
We do this with a large internal tile. Having this work is the first step in improving kiln's ability to lint our tile repos more effectively.

I would appreciate feedback on the user experience and code review :)

To try this out locally
- clone kiln
- run ./install.sh
- cd to some large tile that has templating (ahem p-runtime)
- run something like `kiln pre-process --name ert --tile-names=ert,srt` or `kiln pre-process -n ert -tn=ert,srt`
- see the tile's files after executing the templates, the instance_groups directory is probably the most interesting

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,16 +16,21 @@ kiln
 kiln helps you build ops manager compatible tiles
 
 Usage: kiln [options] <command> [<args>]
-  --help, -h                                               bool    prints this usage information (default: false)
-  --pivotal-network-token, -pt, PIVOTAL_NETWORK_API_TOKEN  string  uaa access token for network.pivotal.io
-  --version, -v                                            bool    prints the kiln release version (default: false)
+  --help, -h     bool  prints this usage information (default: false)
+  --version, -v  bool  prints the kiln release version (default: false)
 
 Commands:
-  bake     bakes a tile
-  fetch    fetches releases
-  help     prints this usage information
-  update   updates stemcell_criteria and releases
-  version  prints the kiln release version
+  bake                    bakes a tile
+  compile-built-releases  compiles built releases and uploads them
+  fetch                   fetches releases
+  find-release-version    prints a json string of a remote release satisfying the Kilnfile version and stemcell constraints
+  help                    prints this usage information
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -16,16 +16,22 @@ kiln
...
-  fetch    fetches releases
-  help     prints this usage information
-  update   updates stemcell_criteria and releases
-  version  prints the kiln release version
+  bake                    bakes a tile
+  compile-built-releases  compiles built releases and uploads them
+  fetch                   fetches releases
+  find-release-version    prints a json string of a remote release satisfying the Kilnfile version and stemcell constraints
+  help                    prints this usage information
+  pre-process             preprocess yaml files
```
ciriarte, line 28: I appreciate you updating the README 🥳

acceptance/help_test.go
```diff
@@ -22,6 +22,7 @@ Commands:
   fetch                   fetches releases
   find-release-version    prints a json string of a remote release satisfying the Kilnfile version and stemcell constraints
   help                    prints this usage information
+  pre-process             preprocess yaml files
```
ciriarte, line 25: I'm still unsure what pre-process means in this context

ciriarte (commented):
@crhntr I think the code looks reasonable, I still don't really understand what it is doing. Can you share a use case for this change?

acceptance/help_test.go
crhntr: For future readers (Carlos and I talked on zoom), the code kind-of-simply runs the config files through [Go standard library's  template library](https://golang.org/pkg/text/template/). We use this internally on the Platform Provider Team to do conditional logic for a set of config files that are used to generate two different tiles. The "tiles" have different instance group colocation strategies (Tanzu Application Service for VMs small footprint (srt) vs. the full tiles). We have been doing this for a long time and are trying to reduce the complexity of what it takes to build our tiles. So, we are adding proposing kiln has this feature as a first class feature. In the future, I would like kiln bake to handle templated configuration files gracefully.

PR #257, opened 2021-03-26
Bake requires fewer flags

crhntr:
This backwards compatible change sets reasonable defaults for bake flags iff the expected files/directories exist.

This allows reducing the the number of required build flags on non-pre-processed tiles to a more reasonable set.

## Example

For example the bake command used for the IST tile requires well over 10 flags.

With this change, a valid tile can be built like this:
```sh
kiln bake --variables-file variables/ist.yml --variable "metadata-git-sha=$(git rev-parse HEAD)"
```

###  Handling "bake"ing from a different dir

When the `--kilnfile` flags is passed and the path contains a directory prefix, like `kiln bake --kilnfile=some-dir/Kilnfile` all defaults are prefixed with the Kilnfile dir. This makes it easy to bake from another directory when metadata directories are where they are expected Relative to where the kilnfile should be. So...
```
kiln bake --kilnfile=some-dir/Kilnfile
```
is equivalent to
```
kiln bake --kilnfile=some-dir/Kilnfile --metadata=some-dir/base.yml
```
when "base.yml" is in the same directory as "Kilnfile".

This change is motivated by how the example tile is baked in CI.

## Not Implemented
- metadata directory config in Kilnfile => we decided against this because Kiln we believe should be opinionated. This change encourages users to use default directories
- logging => I did not log when a default was used or when the flag value was used. This might be helpful for CI debugging.
- default value for "--variables-file" => since in p-runtime we have a single variables directory for both ert and ist, a default would have to figure out which one of those to use.

## Code review notes

I used reflection. This makes it so the default values are specified in one place (the Kiln.Options struct tags). It is a bit weird but it does make it simple to change default values.


commands/bake_internal_test.go
```diff
@@ -0,0 +1,256 @@
...
+				return []byte("4.2.0"), nil
+			default:
+				return nil, os.ErrNotExist
+			}
+		}
+	)
+
+	err := bake.loadFlags([]string{}, statNoError, readFile)
+
+	please.Expect(err).NotTo(Ω.HaveOccurred())
```
ciriarte, line 32: I like how this reads

ciriarte (approved):
## 🥳 Praise 

This is one of the best pull requests I've seen to kiln in a while. This is awesome!

## 🤓 Minor corrections (non blockers):

there's a typo in this sentence (I only care because the message in the PR is part of the commit message):

>kiln bake --kilnfile=some-dir/Kilnfile --meatadata=some-dir/base.yml

and here:

> metatdata directory config in Kilnfile => we decided against this because Kiln we believe should be opinionated. This change encourages users to use default directories

## 🤔  Thoughts

- I think we can achieve `kiln bake` by treating SF-TAS (srt) as a variant and keeping a true "base".
- I would prefer smaller pull requests to separate preprocess from the simplified bake calls.

ciriarte:
Awesome, I'd say let's merge it.

PR #263, opened 2021-07-12
feat: `kiln validate` checks for valid yaml and release version constaint syntax

crhntr:
This implementation checks for valid Kilnfile and Kilnfile.lock YAML syntax. It also ensures the release constraint syntax is valid.


commands/validate.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/pivotal-cf/jhanda"
+	"github.com/pivotal-cf/kiln/internal/cargo"
+	"gopkg.in/src-d/go-billy.v4"
+	"gopkg.in/src-d/go-billy.v4/osfs"
+	"strings"
+)
+
+type Validate struct {
+	Options struct {
+		Kilnfile       string   `short:"kf" long:"kilnfile" default:"Kilnfile" description:"path to Kilnfile"`
```
ciriarte, line 15: **suggestion:** I like examples in my help descriptions.

commands/validate.go
```diff
@@ -0,0 +1,91 @@
...
+		VariablesFiles []string `short:"vf" long:"variables-file" description:"path to variables file"`
```
ciriarte, line 16: **question:** do we have examples of var files?

ciriarte (approved):
Beautiful work.

**suggestion:** let's update the README with the new command.

commands/validate.go
crhntr: Maybe we can create an issue for this so we do it across all the flags.

commands/validate.go
crhntr: Yes. Kiln fetch uses them for s3 credentials. The README has an example.

PR #270, opened 2021-07-23
feat: Added new find-stemcell-version command and modified update-stemcell command

ram-pivot:


commands/find_stemcell_version.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	var productSlug=""
+
+	//Get Stemcell OS and major from Kilnfile
+	if kilnfile.Stemcell.OS == "ubuntu-xenial" {
+		productSlug = "stemcells-ubuntu-xenial"
+	} else if kilnfile.Stemcell.OS == "windows2019" {
+		productSlug = "stemcells-windows-server"
+	}
```
crhntr, line 60: Since we already have the idea of a product slug in the Kilnfile. I think it might be easy enough to edit the Kilnfile stemcell section to specify the slug in addition to the name and version constraint.

commands/find_stemcell_version.go
```diff
@@ -0,0 +1,138 @@
...
+
+    if productSlug == "" {
+		return fmt.Errorf(ErrStemcellOSInfoMustBeValid)
+	} else if kilnfile.Stemcell.Version == "" {
```
crhntr, line 64: I find it easier to follow code when the happy path is not indented. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

commands/find_stemcell_version_test.go
```diff
@@ -0,0 +1,243 @@
...
+					Expect(executeErr).NotTo(HaveOccurred())
+					Expect((&writer).String()).To(ContainSubstring("\"456.118\""))
+					Expect((&writer).String()).To(ContainSubstring("\"remote_path\":\"network.pivotal.io\""))
+					Expect((&writer).String()).To(ContainSubstring("\"source\":\"Tanzunet\""))
+				})
+			})
+		})
+	})
+
+	Describe("Execute", func() {
```
crhntr, line 159: This should probably be named "ExtractMajorVersion" the previous `Describe` is testing "Execute".

commands/update_stemcell.go
```diff
@@ -3,21 +3,21 @@ package commands
...
 	"log"
+	"strings"
 
 	"github.com/pivotal-cf/jhanda"
-	"github.com/pivotal-cf/kiln/builder"
 	"github.com/pivotal-cf/kiln/fetcher"
-	"github.com/pivotal-cf/kiln/helper"
 	"github.com/pivotal-cf/kiln/release"
+	"github.com/Masterminds/semver"
 	"gopkg.in/src-d/go-billy.v4/osfs"
```
crhntr, line 12: (feel free to ignore this comment)

I usually sort my import statements into these ordered blocks (based on [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) but extended a little bit)
- standard library
- dot imports (https://github.com/golang/go/wiki/CodeReviewComments#import-dot)
- test helper/matcher packages
- external libraries
- packages in the current module

fetcher/pivnet.go
```diff
@@ -27,58 +28,68 @@ type Pivnet struct {
...
+	if majorStemcellVersion == "" {
+		return "", ErrStemcellMajorVersionMustNotBeEmpty
 	}
+
 	locator := url.URL{
 		Scheme: "https",
 		Host:   "network.pivotal.io",
-		Path:   path.Join("/api/v2/products", string(slug), "releases"),
+		Path:   fmt.Sprintf("/api/v2/products/%s/releases/latest?major=%s", slug, majorStemcellVersion),
+	}
```
crhntr, line 52: 💯   Nice! I like using `url.URL`. It is uncommon but I feel it is easier to follow than a really long `fmt.Sprintf`.

The RawQuery is a bit oddly named but it would be where you put the query for this URL instead of appending to Path.

commands/find_stemcell_version.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		//Get stemcell version from pivnet
+		stemcellVersion, err := cmd.pivnetService.StemcellVersion(productSlug, majorVersion)
+
+		if err != nil {
+			return err
+		}
+
+		stemcellVersionJson, _ := json.Marshal(stemcellVersionOutput{
```
crhntr, line 81: I would at least return the marshal error. It is rare that something would go wrong but ignoring errors should be a rare edge case. I would not write a test for such an error check.

commands/find_stemcell_version.go
```diff
@@ -0,0 +1,138 @@
...
+			Version: stemcellVersion,
+			Source: "Tanzunet",
+			RemotePath: TanzunetRemotePath,
+		})
+		cmd.outLogger.Println(string(stemcellVersionJson))
+	}
+	return nil
+}
+
+func ExtractMajorVersion(version string) (string, error) {
```
crhntr, line 91: (this is a really loosly held opinion... feel free to ignore it)

Version constraints from this package may not match what the code below seems to expect. It might make sense to fetch the versions from Pivnet and then see which version matches the constraint. Another option would be to instead of specifying the version have the key specify the version line. For example:

```go
package cargo

type StemcellConstraint struct {
	Slug         string `yaml:"slug"`
	MajorVersion int    `yaml:"version_line"`
}
```

crhntr (changes requested):
Looks solid for the most part 👍 

I think it might be good to specify the slug in the Kilnfile's stemcell_critera section instead of having control flow in the command.  For example you can extend the type like this:

```go
package cargo

//...

type PublishedStemcell struct {
	Stemcell
	ProductSlug string `yaml:"slug"`
}
```
Then use this variant in the Kilnfile struct.

In addition to the review comments. These are a few small things I noticed:
- The Go code could use a `go fmt ./...`.
- My linter is warning about error messages starting with an upper case. I kind of like how they are currently and would not mind merging them with upper case letters. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

commands/find_stemcell_version.go
ram-pivot: Yhea I keep forgetting the go fmt ./... thanks for noticing that. Also as discussed I am not making the product slug changes for now.

commands/find_stemcell_version.go
ram-pivot: Thanks for the link. Yhea I agree I changed this one.

commands/find_stemcell_version_test.go
ram-pivot: Yhea this one was a typo 👍

commands/update_stemcell.go
ram-pivot: Yhea I liked this. Usually I run tools for other languages that does this for me. Probably I need to setup something for golang.

For now I manually updated the imports.

fetcher/pivnet.go
ram-pivot: That is cool to know. I added the RawQuery.  I ran into issue with the query parameter adding additional / to the value. So have to use fmt.Sprintf.

commands/find_stemcell_version.go
ram-pivot: This one makes sense. I am returning the error now.

commands/find_stemcell_version.go
ram-pivot: Yhea for now the Kilnfile has options to pin a version. May be we can later expand if we want robust semver rules. I assumed for Stemcell we might not have very complex rules and just limited based on the major alone. But good point where we might have to consider for complex version constraints.

ram-pivot:
@crhntr When you get a chance please check and let me know if anything else needs to be updated. Thanks again for taking the time to review the PR.

crhntr (approved):

PR #283, opened 2021-10-01
feat: stricter Kilnfile validation - release version in lockfile must be valid and match constraint

crhntr:
When this is not true, a descriptive error is output. This change will make it more likely that at any given time, the repository is fetch-able and bake-able.

## Contributor Impact
This will require contributors to run `kiln update-release` when unpinning. This will make it so we reduce the number of PRs and simplify the process of getting a release note for unlocking releases.

### DIFF
--- a/internal/commands/validate.go
+++ b/internal/commands/validate.go
@@ -7,6 +7,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/go-git/go-billy/v5"
 	"github.com/pivotal-cf/jhanda"
+
 	"github.com/pivotal-cf/kiln/internal/commands/flags"
 	"github.com/pivotal-cf/kiln/pkg/cargo"
 )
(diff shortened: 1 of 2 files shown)


ciriarte:
@crhntr ideally, I'd like to see our bot detect and correct this issue in the contextual pull request.

I think we discussed this issue briefly, @crhntr. Happy to revisit it.

ciriarte (approved):
@crhntr ideally, I'd like to see our bot detect and correct this issue in the contextual pull request.

I think we discussed this issue briefly, @crhntr. Happy to revisit it.

e.g. Dependabot or some automation would do something like:

Detected version mismatch.
Commit fixing version mismatch (in the same PR where the Kilnfile was changed)
CI trigger with full validation.

crhntr:
Yup I remember that conversation. I like that workflow too. I'm happy to work towards that.

PR #288, opened 2021-10-27
refactor: release sources

crhntr:
We have a bit of work upcoming to add/refactor/use release sources.

The main changes are
- merging the `internal/fetcher` and `pkg/release` package into `internal/component`
- deduplicating similar data structures `internal/component` and `pkg/cargo`
- documenting and simplifying the interfaces in `internal/component`
  - set download threads once for the s3 release source (instead of every download)
- updating existing commands to use the new interfaces


internal/commands/find_release_version.go
```diff
@@ -3,11 +3,11 @@ package commands
 import (
 	"encoding/json"
 	"github.com/pivotal-cf/kiln/internal/commands/flags"
+	"github.com/pivotal-cf/kiln/internal/component"
```
rhall-pivotal, line 6: Fix yo imports @crhntr :)

internal/component/bosh_io_release_source.go
```diff
@@ -126,52 +124,52 @@ func (src BOSHIOReleaseSource) FindReleaseVersion(requirement release.Requiremen
...
 				return builtRelease, true, nil
 			}
 		}
 	}
-	return release.Remote{}, false, nil
+	return Lock{}, false, nil
 }
 
-func (src BOSHIOReleaseSource) DownloadRelease(releaseDir string, remoteRelease release.Remote, downloadThreads int) (release.Local, error) {
+func (src BOSHIOReleaseSource) DownloadRelease(releaseDir string, remoteRelease Lock, _ int) (Local, error) {
```
rhall-pivotal, line 135: Let take a look at what's going on here

internal/component/release_source.go
```diff
@@ -0,0 +1,91 @@
+package component
```
rhall-pivotal, line 1: lets run the formatter on this page

rhall-pivotal (changes requested):

internal/component/bosh_io_release_source.go
crhntr: fixed: https://github.com/pivotal-cf/kiln/pull/288/commits/2707dc33e2d2dc2c1a4abcf4e6c6a40578de6e08

PR #290, opened 2021-10-29
refactor: higher level abstraction for Kiln commands

crhntr:
Currently changes to kilnfile parsing are hard to make.
Also we have many duplicate tests for parsing the kilnfiles.
This refactor does not aim to update the tests to use the new signature but does update most commands to use it and then calls the new helper to return a type that implements `jhanda.Command`.

There are two new interfaces a command may implement either:

```go
// kilnfileLockReturner can be implemented for commands that only read from the kilnfiles
type kilnfileReciever interface {
	KilnExecute(args []string, fn OptionsParseFunc) error
}

// kilnfileLockReturner can be implemented for commands that update the kilnfile.lock
type kilnfileLockReturner interface {
	KilnExecute(args []string, fn OptionsParseFunc) (_ cargo.KilnfileLock, updateFile bool, _ error)
}

// OptionsParseFunc is added here for reference
type OptionsParseFunc func(args []string, opts options.StandardOptionsEmbedder) (cargo.Kilnfile, cargo.KilnfileLock, error)
```
and a type to wrap them `commands.Kiln`. This command abstraction will make testing and adding new commands with a standard way of parsing kilnfile flags and reduce the testing burden on new commands.

Here is a simple example command to print out the product slug.

```go
package commands

import (
	"fmt"
	"io"

	"github.com/pivotal-cf/kiln/internal/commands/options"
	"github.com/pivotal-cf/kiln/pkg/cargo"
)

type GetSlug struct {
	Options struct {
		options.Standard
	}
	
	W io.Writer
}

func NewGetSlug(w io.Writer) GetSlug { return GetSlug{W: w} }

func (g GetSlug) KilnExecute(args []string, parseOpts commands.OptionsParseFunc) error {
	kf, _, _, err := parseOpts(args, &g.Options)
	if err != nil {
		return err
	}
	_, _ = fmt.Fprintf(g.W, kf.Slug)
	return nil
}

func (GetSlug) Usage() jhanda.Usage { return jhanda.Usage{} }

// existing commands can continue to implement `jhanda.Command` (so no tests should change)
// by writing a method like this:

func (g GetSlug) Execute(args []string) error {
	return commands.Kiln{
		Wrapped: g,
	}.Execute(args)
}

```

[This commit](https://github.com/pivotal-cf/kiln/pull/290/commits/f0b4bbd5e801f00b41451731630af67f655cdc21) shows how this improves testing commands and allows them to more clearly test only what they need.


crhntr:
I am not going to update all the tests to use the updated commands.Kiln signature. We can do this over time.

internal/commands/commands.go
```diff
@@ -0,0 +1,195 @@
...
+	"github.com/go-git/go-billy/v5/osfs"
+	"github.com/pivotal-cf/jhanda"
+	"github.com/pivotal-cf/kiln/internal/baking"
+	"github.com/pivotal-cf/kiln/internal/commands/options"
+	"github.com/pivotal-cf/kiln/pkg/cargo"
+)
+
+type OptionsParseFunc func(args []string, opts options.StandardOptionsEmbedder) (cargo.Kilnfile, cargo.KilnfileLock, []string, error)
+
+type kilnfileReciever interface {
```
ciriarte, line 21: **nitpick**: I think this is a typo kilnfileReciever => kilnfileReceiver

ciriarte (approved):
**thought**: I like this change. However, I keep pondering about our need to use `jhanda`. I feel it keeps us from having a lot of niceties in kiln, and we keep iterating on flag parsing. It would be awesome if we could use something like cobra in the future so that we can piggyback on its functionality.

**chore**: Let's fix the conflicts and we're ready to go.

crhntr:
I'd like to hold off on more work until after Go 1.18. I think type parameters might be useful in this implementation.

crhntr:
I think we might have a better way to do this with type parameters in a few months.

PR #295, opened 2021-11-06
bugfix/cache compiled releases/isolate release source

ciriarte:
- fix: add preerelease support
- debug: add tracing to matching logic

### DIFF
--- a/internal/component/s3_release_source.go
+++ b/internal/component/s3_release_source.go
@@ -256,7 +256,7 @@ func (src S3ReleaseSource) UploadRelease(spec Spec, file io.Reader) (Lock, error
 		return Lock{}, err
 	}
 
-	src.logger.Printf("uploading release %q to %s at %q...\n", spec.Name, src.ID(), remotePath)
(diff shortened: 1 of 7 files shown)


rhall-pivotal (approved):
Looks good to me!

rhall-pivotal (approved):

internal/component/s3_release_source.go
```diff
@@ -256,7 +256,7 @@ func (src S3ReleaseSource) UploadRelease(spec Spec, file io.Reader) (Lock, error
 		return Lock{}, err
 	}
 
-	src.logger.Printf("uploading release %q to %s at %q...\n", spec.Name, src.ID(), remotePath)
+	src.logger.Printf("uploading release %q to %s at %q...\n", spec.Name, src.ReleaseSourceConfig.Bucket, remotePath)
```
crhntr, line 259: ID should return the bucket name when the ID is not set. We can merge this now, but I think there is something wrong with the s3 release source implementation.

crhntr (approved):

PR #296, opened 2021-11-09
refactor: merge `component.Spec` and `component.Requirment` and undo embedding `component.Spec` in `component.Lock`

crhntr:
## Caveating the big diff

Most of the diff comes from undoing embedding the `ComponentSpec` into `ComponentLock`. I initially thought embedding would dry the code and at a high level it made sense. The lock would would contain the exact spec that led to resolving to the exact version that was found. But it turns out, it cluttered every usage. The embedding was nice at first glance where the types were defined (if you don't account for all the usages) but once we added a slice it caused an issue with comparability (discussed below). **This is a good reminder of why in Go, the procedures should drive the data structures not the other way around.** Many of the changed files if compared with files from a few weeks back, should only contain identifier name changes. _I intend to no longer restructure the component source data types after this... it's f*ing brutal._ I am thinking about how to make the tests more readable and also easier to refactor, the chain-able methods on `cargo.ComponentLock` and `cargo.ComponentSpec` help but I still think the tests could be improved... but that is for another commit.

## How does this PR improve the component source data types?

Now there is one fewer types embedding the release Name and Version. Instead of having two types, `cargo.ComponentLock` and `component.Requirement` the fields not found in the lock have been added and the `component.Requirement` type is removed. This makes it easier to add additional fields in the Kilnfile and pass those into the release sources. One reason we want to do this is so we can add a repositories field to each release in the Kilnfile in support of future SBOM work and to make implementing the GitHub release source easier.

This change makes the type constraints of the two cargo types a bit clearer. The `ComponentLock` must be comparable and may not contain composite fields. The `ComponentSpec` does not have that requirement. This will make breaking code by adding fields to Kilnfiles less likely.

Although the usage of the types is driving this refactor, if you'd like to see how the types have changed here [they are](https://github.com/pivotal-cf/kiln/pull/296/files#diff-2989cd5c93be0f9d695b8edcc6d15d02876f250f2181b9cee9466cca368112e7). To see how the merging of `component.Requirement` and `component.Spec` turned out [see the `component.S3ReleaseSource`](https://github.com/pivotal-cf/kiln/pull/296/files#diff-37c18d47fedd41f6fa85979c702009f82b82bfe2583c7ac94107dad722cc81c8).

### Undo the Embedding
```diff
  Expect(lock).To(Equal(cargo.ComponentLock{
- 	ComponentSpec: cargo.ComponentSpec{
- 		Name:    "release-a",
- 		Version: "1.2.3",
- 	},
+	Name:         "release-a",
+	Version:      "1.2.3",
	RemoteSource: compiledReleasesID,
	RemotePath:   "release-a/release-a-1.2.3-ubuntu-trusty-22.tgz",
	SHA1:         releaseASha1,
  })
```

### Merging `component.Spec` and `component.Requirment`

The `ComponentSpec` is passed to `compoent.ReleaseSource` implementations ([see](https://github.com/pivotal-cf/kiln/pull/296/files#diff-05bae9a675ad6e73ebadf857d14be8c7872cd7c5862cdd9ed4f9e3b77d09a81fR47-L64)). This allows adding fields to the spec like the following (pulled from the feature branch driving this change) and then using them in the `ReleaseSource` implementations.

```diff
package cargo

type ComponentSpec struct {
	// Name is a required field and must be set with the bosh release name
	Name string `yaml:"name"`
	// Version if not set, it will default to ">0".
	// See https://github.com/Masterminds/semver for syntax
	Version string `yaml:"version,omitempty"`
	// StemcellOS may be set when a specifying a component
	// compiled with a particular stemcell. Usually you should
	// also set StemcellVersion when setting this field.
	StemcellOS string `yaml:"os,omitempty"`
	// StemcellVersion may be set when a specifying a component
	// compiled with a particular stemcell. Usually you should
	// also set StemcellOS when setting this field.
	StemcellVersion string `yaml:"stemcell_version,omitempty"`

+	// Repositories are where the BOSH release source code is
+	Repositories []string `yaml:"repositories,omitempty"`
}
```

## Additional dev notes from the commit

- also use component.Lock as comparable key in maps and switches
  this allows us to add arrays and maybe maps to the spec
- the spec now allows adding fields that do not make sense in the
  Kilnfile such as StemcellOS and StemcellVersion. It might make
  sense to embed cargo.ComponentSpec in comonent.Spec and st the
  stemcell information from the stemcell criteria field
- this also moves the constraint parsing to the cargo package
  this surfaces how the parsing occurs and moves the empty string
  handling closer to the documentation (see how we use ">0"

- this change also includes more documentation various struct fields
  - cargo.ComponentSpec.Name
  - cargo.ComponentSpec.Version
  - cargo.ComponentSpec.StemcellOS
  - cargo.ComponentSpec.StemcellVersion
  - the Repositories field is commented out and has incomplete docs
- the cargo types now have helper chainable methods. These do not modify
  the reciever but use pass-by-value semantics to copy existing fields
  to a new struct and set the expected fields. This functionality is not
  tested but should be.
- this refactor caught the regression found by the folks working on the compiled release source... kind of neat


crhntr:
Oops! I forgot to do a conventional commit. This is a "refactor" commit.

PR #303, opened 2021-12-02
fix: bake fails in tmp dir

crhntr:
The constructor for the ReleaseNotes command expected to be called when the current directory is in a git repo. This is not always true and some existing scripts expect to run kiln from a temporary directory. To not break those scripts, the repository field on release notes is not set in the Execute method and not the constructor.

When testing the issue, I noticed the file name is a bit weird when there is no version set. So I also made the default output file name not have a dash when the version is not configured.

### DIFF
--- a/internal/commands/bake.go
+++ b/internal/commands/bake.go
@@ -215,7 +215,10 @@ func (b *Bake) loadFlags(args []string, stat flags.StatFunc, readFile func(strin
 	}
 
 	if shouldGenerateTileFileName(b, args) {
-		b.Options.OutputFile = "tile-" + b.Options.Version + ".pivotal"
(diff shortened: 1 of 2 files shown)


ciriarte:
@crhntr I'm a bit confused. Is this related to the git repo issue we discussed during tile testing or not? I'm trying to map how release notes affect testing.

crhntr:
The constructor (always called in main) for the ReleaseNotes command expected it to run from a repo.

crhntr:
After merging this you should be able to push a new pre-release tag and use that in the test script.

ciriarte:
Gotcha. It makes sense.

ciriarte (approved):
Thank you for the clarification, it looks great.

PR #305, opened 2021-12-13
Make stemcell version validations stricter

crhntr:
This set of new validations ensures:
- stemcell version lock matches spec constraint
- stemcell version and constraint are valid
- each release lock with a stemcell has the the same os and version as the lock

### DIFF
--- a/pkg/cargo/validate.go
+++ b/pkg/cargo/validate.go
@@ -73,3 +73,42 @@ func checkComponentVersionsAndConstraint(spec BOSHReleaseTarballSpecification, l
 
 	return nil
 }
+
+func checkStemcell(spec Kilnfile, lock KilnfileLock) []error {
+	v, err := semver.NewVersion(lock.Stemcell.Version)
+	if err != nil {
+		return []error{fmt.Errorf("invalid lock version %q in Kilnfile.lock: %w",
+			lock.Stemcell.Version, err)}
+	}
+
+	if spec.Stemcell.Version != "" {
+		c, err := semver.NewConstraint(spec.Stemcell.Version)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


crhntr (commented):
I self-reviewed this a few months after writing it and still feel this is an okay improvement.

pkg/cargo/kilnfile_validate.go
```diff
@@ -57,3 +57,42 @@ func checkComponentVersionsAndConstraint(spec ComponentSpec, lock ComponentLock,
...
+		matches, errs := c.Validate(v)
+		if !matches {
+			return []error{fmt.Errorf("stemcell version %s in Kilnfile.lock does not match constraint %q: %v",
+				lock.Stemcell.Version, spec.Stemcell.Version, errs)}
+		}
+	}
+
+	var result []error
+	for index, componentLock := range lock.Releases {
+		if componentLock.StemcellOS == "" {
```
ram-pivot, line 84: If the kilnfile.lock contains un-compiled releases this function will not catch those? Ideally we don't want un-compiled releases to pass the PR checks. 

In middle of processing or when the PR was originally created having un-compiled releases is fine. But the final stage before merging the PR we need to validate to make sure we don't have any un-compiled releases. 

Other than that changes look good to me. This will prevent bad kilnfile.lock changes to be merged in. Thanks for the changes.

crhntr:
I rebased this branch. I don't know if we even want this functionality any more.

pkg/cargo/kilnfile_validate.go
crhntr: I think the rule around all the releases being compiled could be an additional flag. Mixed compiled and uncompiled releases should pass the basic `kiln validate`. Maybe we should have a name for an addition an additional validation level.

PR #306, opened 2022-01-14
refactor: remove `compile-built-releases`

crhntr:
Similar functionality is now handled by cache-compiled-releases. This is no longer used in our pipelines so we should remove it.


dtimm (approved):
LGTM

PR #308, opened 2022-01-19
refactor: use bosh.io release index file instead of guessing

crhntr:
BOSH releases on Bosh.io's are based on a publicly available static file. Instead of guessing many combinations this change uses the index and reduces the number of API calls significantly.

This change also exposes the component package so we can use it in non-kiln code. This might be better as separate PR but I thought it made sense to put them together as I would like to use this in PPE specific TAS code.


rizwanreza:
@crhntr What impact does it have to a Kiln user or Kilnfile?

Apart from performance, what value do we get from using the index?

crhntr:
It should not adversely affect users of Kiln nor require Kilnfile changes. I just made a small change after reading your question. It now iterates though all the matches (like the old one) but does so without checking a bunch of URLs that don't exist. The new source of truth is the same as the bosh.io website.

One issue both this and the previous implementation have is they cannot disambiguate between:
repo: cloudfoundry/hello-release, release version: v1, digest: blablabla
repo: pivotal/hello-release, release version: v1, digest greetingsgreetings
This might could be a supply chain risk.

The real reason I pushed this is, I had some of this code on my machine for a while and today/yesterday I wanted to practice using github workflows. This branch seemed like an interesting place to get a little more confidence since keeping the cache up to date requires using a different trigger and committing back to the repo.

internal/component/bosh_io_index.yml
```diff
@@ -0,0 +1,390 @@
...
+  url: https://github.com/cloudfoundry/cf-riak-cs-release
+- name: garden-linux
+  url: https://github.com/cloudfoundry-incubator/garden-linux-release
+- name: docker
+  url: https://github.com/cloudfoundry-incubator/docker-boshrelease
+- name: mesos
+  url: https://github.com/cf-platform-eng/mesos-boshrelease
+- name: shipyard
+  url: https://github.com/cf-platform-eng/shipyard-boshrelease
+- name: ""
```
staylor14, line 173: Cut-and-paste leftover?

staylor14 (commented):
I can't comment on lines that haven't changed, so I have to do it here.

In NewBOSHIOReleaseSource, why do we test c.Type is not an empty string?  Shouldn't we just require that it have the value `ReleaseSourceTypeBOSHIO`?

internal/component/bosh_io_index.yml
crhntr: I agree it is weird that we are testing for a non-empty string. It should be ReleaseSourceTypeBOSHIO whenever NewBOSHIOReleaseSource is called.

ciriarte (commented):
**question**: this looks great, but let me see if I understand: it looks like this index is created at compile-time and embedded into the release?

What are the risks associated with hitting a stale index? Does it fall back to the previous behavior? From what I read, it looks like we will need to recompile and publish a new version.

crhntr:
> **question**: this looks great, but let me see if I understand: it looks like this index is created at compile-time and embedded into the release?
> 
> What are the risks associated with hitting a stale index? Does it fall back to the previous behavior? From what I read, it looks like we will need to recompile and publish a new version if the index is stale.

Yes it would need to be recompiled. It should fallback.

crhntr:
After a few years without being merged. I think it is time this PR is closed.

PR #310, opened 2022-02-04
Add darwin_arm64 builds to goreleaser

dtimm:
This will add automatic builds for Apple computers with M1 architecture to published releases.

It changes the names of build artifacts to include architecture, as well.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -10,22 +10,21 @@ release:
 builds:
 - env:
   - CGO_ENABLED=0
-  goos:
-  - darwin
-  - linux
-  - windows
-  goarch:
-  - amd64
+  targets:
+  - linux_amd64
+  - windows_amd64
+  - darwin_amd64
+  - darwin_arm64
   ldflags:
   - -X "main.version={{.Version}}"
   main: ./
 archives:
   - id: github
     format: binary
-    name_template: "kiln-{{ .Os }}-{{ .Version }}"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -10,22 +10,21 @@ release:
 builds:
 - env:
   - CGO_ENABLED=0
-  goos:
-  - darwin
-  - linux
-  - windows
-  goarch:
-  - amd64
+  targets:
```
crhntr, line 13: I like this list. It is easier to follow than the matrix based config.

.goreleaser.yml
```diff
@@ -10,22 +10,21 @@ release:
...
+  - darwin_amd64
+  - darwin_arm64
   ldflags:
   - -X "main.version={{.Version}}"
   main: ./
 archives:
   - id: github
     format: binary
-    name_template: "kiln-{{ .Os }}-{{ .Version }}"
+    name_template: "kiln-{{ .Os }}-{{ .Arch }}-{{ .Version }}"
```
crhntr, line 24: I wonder if this will break any consumers? I guess we will need to see. Although this looks like it is using Go (stdlib) templates, I don't think we want to do any "if" stuff here for backwards compatibility. I think this is fine. If things break I am okay being on call for this one.

crhntr (approved):
Looks great! Thanks :)

PR #311, opened 2022-02-22
expose "historic" package

crhntr:
feat [pkg]: move historic package from internal
this will help us reconstruct the BOM from tiles on pivnet


jaristiz (commented):
Looks good.

PR #312, opened 2022-03-08
feature: add metadata validation

crhntr:


crhntr:
Won't do. @staylor14 and I think this would be better served by exporting some sort of yaml schema from Ops Manager and using that to validate the metadata.yml. The proofing package is probably too much work to maintain and keep up to date with Ops Manager.

PR #313, opened 2022-03-25
feat: use only successfully configured release sources

crhntr:
**The goal is to make additions/deletions from the Kilnfile lock release_sources section not fail in CI unless they absolutely have to.**

The high-level change to release source configuration is here: https://github.com/pivotal-cf/kiln/blob/a8d282c2735d73cd2646f7ee4f06c9352b14e369/internal/commands/flags/parse.go#L78-L109

To fix the issue of kiln failing when new releases sources are added but the configuration for those release sources is incomplete, I made the Kilnfile loader log the misconfigured release sources but still continue on. I think this will work with most commands. It has the benefit of notifying contributors to add configuration when needed but also keeps backwards compatibility when we add new release sources and contributors have not yet added credentials.

We will be adding an Artifactory release source shortly so this might help with that.

The refactors included in the PR attempt to 
- split up and remove the cargo.ReleaseSourceSpec type
- move the string field manipulations (parsing, validation, fallbacks, defaults...) to one place, the cargo package.
- separate concerns for different release sources to separate types. The names of the structs in "github.com/pivotal-cf/kiln/pkg/cargo" (for example S3ReleaseSource) align with the implementations in "github.com/pivotal-cf/kiln/internal/component". The former deals with config the latter implementation.
- outline a pattern for adding configurations for new release sources
- scratch my need-to-refactor itch


crhntr:
I just pushed a commit where I squashed the fixes with the "feat" commit.

crhntr:
@staylor14 @rhall-pivotal The last commit could be reverted as I don't think that is the behavior we want moving forward. I argue the refactoring commits that come before it are valuable and would like to see them in a separate PR if this change is rejected.

crhntr:
Won't do. We are moving forward with a bigger refactor of release sources.

PR #307, opened 2022-01-14
refactor: fix Goland linter warnings and run stricter formatter (gofumpt)

crhntr:
I ran the Goland code analysis tool and fixed all the "Go" warnings.

PR #317, opened 2022-04-26
refactor: apply stricter formatting rules

crhntr:
these are optional but are kind of nice.

I don't think we need to add additional tooling to do this but I thought I'de run it as a one off fix.


crhntr:
already merged in #307

PR #318, opened 2022-04-26
bump (go-pivnet): import v7 everywhere

crhntr:
dependabot is not bumping major versions so I am doing it manually


crhntr:
I just rebased the PR on main.

crhntr:
Those were probably part of old code that has since been removed. This PR is really old.

crhntr:
> Those were probably part of old code that has since been removed. This PR is really old. I'll clean it up.

Looks like the generate statements were a bit off. I fixed them and ran generate again. This should be good now. I guess the find and replace for `interface{}` -> `any` was a bit aggressive. I should not have changed the contents of generated files.

pabloarodas (approved):
Looks good to me now

PR #322, opened 2022-07-06
refactor: rename package historic to history

crhntr:
the package interacts with history it is not itself a thing from the past


staylor14 (approved):
This warms my cold dead heart!

PR #325, opened 2022-07-28
feature (release-notes): parse version from tag suffix instead of from VERSION file

crhntr:
Currently the command requires the final revision to contain a version file.
This PR adds fall back behavior to read the version from the final revision (aka tag) if it exists. For example it would parse the final version from `ist/2.13.2` properly.


crhntr:
It probably makes more sense to use a flag on the command.

PR #333, opened 2022-09-21
Release specifications in the Kilnfile require a release_source

crhntr:
Releases now must specify a release_source property like so and kiln will only check the specified release_source for a release:

```diff
 ---
 release_sources:
   - type: s3
     bucket: some-bucket
 releases:
   - name: some-offline-buildpack
+    release_source: some-bucket
```

## TODO
- [x] Make release source required
- [x] Check release target for already cached release before upload
- [x] Add check in validate to ensure release_source field is populated


crhntr:
I think a re-design of how we do release consumption is a better use of time than this PR.

PR #337, opened 2022-10-31
refactor (validate): move validations to cargo package

crhntr:


crhntr:
This is a small change. It just moves some stuff to the Cargo package.

PR #343, opened 2023-02-07
log to stderr to not pollute the output

syslxg:
Co-authored-by: Gary Liu <garyliu@vmware.com>


crhntr:
I like this change!

crhntr (approved):
Looks good!

PR #344, opened 2023-02-13
update usage message for fetch cmd

syslxg:
Co-authored-by: Gary Liu <garyliu@vmware.com>


crhntr (approved):
Looks great! Thanks :)

I might suggest we differentiate "BOSH releases" from "releases" in some future change.

PR #351, opened 2023-03-06
Export functions for baking and tile metadata generation

crhntr:
## Why re-write Bake?

Kiln Bake the CLI is now pretty easy to invoke. Few flags are actually required. Calling Bake from Go is not easy. It would be nice if the Kiln Bake code could be easily be called as a library and if it was easier to understand. This codebase is an attempt to rewrite bake command to make it easier to import into other Go codebases.


crhntr:
i still think i want to do this but it should be done from a more recent starting point

PR #358, opened 2023-03-09
refactor: add more static analysis

crhntr:
I added some of the linters form this page: https://golangci-lint.run/usage/linters/#disabled-by-default

Each linter is added as a separate commit so the changes to code can be tied to a particular linter that forced the change.

Some linters did not require a change but I added them because I remember a time when I've mentioned the fix they suggest during pairing.


crhntr:
When I created this PR is was un-aware we decided on [staticcheck](https://staticcheck.io/). I will close this PR and create a new one with that tool.

PR #367, opened 2023-03-28
add acceptance tests to test workflow

crhntr:


crhntr:
This should be re-created once main is fixed.

PR #370, opened 2023-04-04
fix: kiln test docker output

notrepo05:
Authored-by: Nick Rohn <nrohn@vmware.com>


crhntr (approved):
As a bug fix. This looks good.

This does not seem like the time of thing we could easily test.

PR #373, opened 2023-04-04
feat: fetch release notes for component bumps from trainstat

pabloarodas:
- Add -tu flag to release notes command to specify a different trainstat url

Authored-by: Pablo Rodas <prodas@vmware.com>

### DIFF
--- a/internal/commands/release_notes.go
+++ b/internal/commands/release_notes.go
@@ -33,6 +33,7 @@ type ReleaseNotes struct {
 		Kilnfile     string `long:"kilnfile"     short:"k" description:"path to Kilnfile"`
 		DocsFile     string `long:"update-docs"  short:"u" description:"path to docs file to update"`
 		notes.IssuesQuery
+		notes.TrainstatQuery
 	}
 
 	repository *git.Repository
@@ -45,7 +46,7 @@ type ReleaseNotes struct {
 	repoOwner, repoName string
 }
 
-type FetchNotesData func(ctx context.Context, repo *git.Repository, client *github.Client, tileRepoOwner, tileRepoName, kilnfilePath, initialRevision, finalRevision string, issuesQuery notes.IssuesQuery) (notes.Data, error)
+type FetchNotesData func(ctx context.Context, repo *git.Repository, client *github.Client, tileRepoOwner, tileRepoName, kilnfilePath, initialRevision, finalRevision string, issuesQuery notes.IssuesQuery, trainstatClient notes.TrainstatFetcher) (notes.Data, error)
 
 func NewReleaseNotesCommand() (ReleaseNotes, error) {
 	return ReleaseNotes{
@@ -86,12 +87,15 @@ func (r ReleaseNotes) Execute(args []string) error {
 		client = gh.Client(ctx, r.Options.GithubToken)
 	}
 
+	var trainstatClient = notes.NewTrainstatClient(r.Options.TrainstatQuery.TrainstatUrl)
+
 	_ = notes.FetchData // fetchNotesData is github.com/pivotal/kiln/internal/notes.FetchData
 	data, err := r.fetchNotesData(ctx,
 		r.repository, client, r.repoOwner, r.repoName,
 		r.Options.Kilnfile,
 		nonFlagArgs[0], nonFlagArgs[1],
 		r.Options.IssuesQuery,
+		&trainstatClient,
 	)
 	if err != nil {
 		return err
--- a/internal/commands/release_notes.go
+++ b/internal/commands/release_notes.go
@@ -46,7 +46,7 @@ type ReleaseNotes struct {
 	repoOwner, repoName string
 }
 
-type FetchNotesData func(ctx context.Context, repo *git.Repository, client *github.Client, tileRepoOwner, tileRepoName, kilnfilePath, initialRevision, finalRevision string, issuesQuery notes.IssuesQuery, trainstatClient notes.TrainstatFetcher) (notes.Data, error)
+type FetchNotesData func(ctx context.Context, repo *git.Repository, client *github.Client, tileRepoOwner, tileRepoName, kilnfilePath, initialRevision, finalRevision string, issuesQuery notes.IssuesQuery, trainstatClient notes.TrainstatNotesFetcher) (notes.Data, error)
 
 func NewReleaseNotesCommand() (ReleaseNotes, error) {
 	return ReleaseNotes{
@@ -87,7 +87,7 @@ func (r ReleaseNotes) Execute(args []string) error {
 		client = gh.Client(ctx, r.Options.GithubToken)
 	}
 
-	var trainstatClient = notes.NewTrainstatClient(r.Options.TrainstatQuery.TrainstatUrl)
+	var trainstatClient = notes.NewTrainstatClient(r.Options.TrainstatQuery.TrainstatURL)
 
 	_ = notes.FetchData // fetchNotesData is github.com/pivotal/kiln/internal/notes.FetchData
 	data, err := r.fetchNotesData(ctx,
--- a/internal/commands/release_notes_test.go
+++ b/internal/commands/release_notes_test.go
@@ -72,7 +72,7 @@ func TestReleaseNotes_Execute(t *testing.T) {
 			repoOwner:  "bunch",
 			repoName:   "banana",
 			readFile:   readFileFunc,
-			fetchNotesData: func(c context.Context, repo *git.Repository, ghc *github.Client, tro, trn, kfp, ir, fr string, iq notes.IssuesQuery) (notes.Data, error) {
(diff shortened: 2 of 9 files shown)


pkg/notes/notes_data.go
```diff
@@ -513,3 +532,74 @@ func setEmptyComponentGitHubRepositoryFromOtherKilnfile(k1, k2 cargo.Kilnfile) c
 }
 
 func closeAndIgnoreError(c io.Closer) { _ = c.Close() }
+
+type TrainstatFetcher interface {
```
crhntr, line 536: I'd call this "TrainstatNotesFetcher" to make it align with the method name better.

pkg/notes/notes_data.go
```diff
@@ -513,3 +532,74 @@ func setEmptyComponentGitHubRepositoryFromOtherKilnfile(k1, k2 cargo.Kilnfile) c
...
+	client := &http.Client{
+		Timeout: time.Second * 10,
+	}
+	return TrainstatClient{
+		host:       host,
+		httpClient: client,
+	}
+}
+
+func (t *TrainstatClient) FetchTrainstatNotes(milestone string, version string, tile string) (notes []string, err error) {
```
crhntr, line 558: FetchTrainstatNotes should receive a context as the first parameter and pass it to http.NewRequestWithContext (instead of http.NewRequest).

pkg/notes/notes_data.go
```diff
@@ -513,3 +532,74 @@ func setEmptyComponentGitHubRepositoryFromOtherKilnfile(k1, k2 cargo.Kilnfile) c
...
+	q.Set("version", version)
+	q.Set("tile", tile)
+	req.URL.RawQuery = q.Encode()
+
+	res, err := t.httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
```
crhntr, line 581: consider using the closeAndIgnoreError helper (or adding it to this package). This makes linters less upset about ignoring this error.

```go
func closeAndIgnoreError(c io.Closer) {
   _ = c.Close()
}
```

pkg/notes/notes_data.go
```diff
@@ -513,3 +532,74 @@ func setEmptyComponentGitHubRepositoryFromOtherKilnfile(k1, k2 cargo.Kilnfile) c
...
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	q := req.URL.Query()
+	q.Set("milestone", milestone)
+	q.Set("version", version)
+	q.Set("tile", tile)
+	req.URL.RawQuery = q.Encode()
+
+	res, err := t.httpClient.Do(req)
```
crhntr, line 576: httpClient will not return an error on non success status codes.

Write a check on req.StatusCode after the Do error check.

pkg/notes/notes_data.go
```diff
@@ -513,3 +532,74 @@ func setEmptyComponentGitHubRepositoryFromOtherKilnfile(k1, k2 cargo.Kilnfile) c
...
+
+	if err = json.Unmarshal(responseData, &notes); err != nil {
+		return notes, err
+	}
+	return
+}
+
+func (t *TrainstatClient) tileSupported(tile string) bool {
+	switch tile {
+	case "elastic-runtime":
```
crhntr, line 596: consider multi value switch statement.

```go
switch tile {
case "a", "b", "c":
  return true
default:
  return false
}
```

crhntr: I would prefer not to have the slugs in the code and for the code to handle a "404" from train stat properly.

crhntr (changes requested):

internal/commands/release_notes_test.go
```diff
@@ -72,7 +72,7 @@ func TestReleaseNotes_Execute(t *testing.T) {
 			repoOwner:  "bunch",
 			repoName:   "banana",
 			readFile:   readFileFunc,
-			fetchNotesData: func(c context.Context, repo *git.Repository, ghc *github.Client, tro, trn, kfp, ir, fr string, iq notes.IssuesQuery) (notes.Data, error) {
+			fetchNotesData: func(c context.Context, repo *git.Repository, ghc *github.Client, tro, trn, kfp, ir, fr string, iq notes.IssuesQuery, tf notes.TrainstatFetcher) (notes.Data, error) {
```
notrepo05, line 75: I might suggest renaming these arguments to be longer, such as using trainstatClient instead of tf, like the above change

pkg/notes/notes_data.go
```diff
@@ -513,3 +532,74 @@ func setEmptyComponentGitHubRepositoryFromOtherKilnfile(k1, k2 cargo.Kilnfile) c
...
+type TrainstatFetcher interface {
+	FetchTrainstatNotes(milestone string, version string, tile string) ([]string, error)
+}
+
+type TrainstatClient struct {
+	host       string
+	httpClient *http.Client
+}
+
+func NewTrainstatClient(host string) TrainstatClient {
```
notrepo05, line 545: is quoted "null" intentional here?

notrepo05 (commented):
Looks good. Echoing what Chris suggested and just added two comments for a double check

internal/commands/release_notes_test.go
crhntr: You might be able to get away with "_" instead of "tf" or "trainstatClient". I don't see any usages of that parameter in this helper function.

pkg/notes/notes_data.go
pabloarodas: Thanks @crhntr , didn't know we could use switch statements like that.

The purpose of this function is to know if we should call trainstat to fetch notes. Trainstat only supports TAS tiles, and for the acceptance tests we are running the fetch release notes logic for the hello-tile.

pkg/notes/notes_data.go
pabloarodas: Done!

pkg/notes/notes_data.go
pabloarodas: Oh, I didn't realize about this earlier, thanks for pointing this out, that condition is not necessary, I'll delete it.

pkg/notes/notes_data.go
pabloarodas: Yes, that makes sense. Done!

pkg/notes/notes_data.go
```diff
@@ -90,6 +92,15 @@ type IssuesQuery struct {
 	IssueTitleExp  string   `long:"issues-title-exp"       short:"x" description:"issues with title matching regular expression will be added. Issues must first be fetched with github-issue* flags. The default expression can be disabled by setting an empty string" default:"(?i)^\\*\\*\\[(security fix|feature|feature improvement|bug fix|breaking change)\\]\\*\\*.*$"`
 }
 
+type TrainstatQuery struct {
+	TrainstatUrl string `long:"trainstat-url" short:"tu" description:"trainstat url to fetch the release notes for component bumps" default:"https://trainstat.sc2-04-pcf1-apps.oc.vmware.com"`
+}
+
+func TrainstatUrl() string {
+	f, _ := reflect.ValueOf(TrainstatQuery{}).Type().FieldByName("TrainstatUrl")
```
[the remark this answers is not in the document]
pabloarodas: Is it necessary to panic? maybe failing the test is enough to make people know that the field name should not change. I wrote a test that fails when the field is not found.

pkg/notes/notes_data.go
```diff
@@ -90,6 +92,15 @@ type IssuesQuery struct {
 	IssueTitleExp  string   `long:"issues-title-exp"       short:"x" description:"issues with title matching regular expression will be added. Issues must first be fetched with github-issue* flags. The default expression can be disabled by setting an empty string" default:"(?i)^\\*\\*\\[(security fix|feature|feature improvement|bug fix|breaking change)\\]\\*\\*.*$"`
 }
 
+type TrainstatQuery struct {
+	TrainstatUrl string `long:"trainstat-url" short:"tu" description:"trainstat url to fetch the release notes for component bumps" default:"https://trainstat.sc2-04-pcf1-apps.oc.vmware.com"`
+}
+
+func TrainstatUrl() string {
```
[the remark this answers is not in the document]
pabloarodas: Done!

internal/commands/release_notes.go
```diff
@@ -86,12 +87,15 @@ func (r ReleaseNotes) Execute(args []string) error {
 		client = gh.Client(ctx, r.Options.GithubToken)
 	}
 
+	var trainstatClient = notes.NewTrainstatClient(r.Options.TrainstatQuery.TrainstatUrl)
```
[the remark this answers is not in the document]
pabloarodas: Done!

internal/commands/release_notes_test.go
pabloarodas: Thank you both, I changed the name to "_" since as Chris pointed out that value is not being used in the function.

notrepo05 (approved):

PR #378, opened 2023-04-14
Feat/test verbose flag

notrepo05:
* Make `parallel` ginkgo testing default
* By default, suppress "Info:" messages in `kiln test`
* Elaboration in docs on `test` and `bake`

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,10 @@ provides_product_versions:
   version: $( version )
 ```
 
+#### `--skip-fetch-directories`
+
+The `--skip-fetch-directories` flag bypasses the default behavior to fetch releases when running `kiln bake`.
+
 </details>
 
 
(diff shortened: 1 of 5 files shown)


notrepo05:
Example of this running

```
➜  kiln git:(feat/test-verbose-flag) ✗ ./kiln test -tp ~/workspace/tas/ist
Running Suite: Manifest Generation Suite
========================================
Random Seed: 1681503414
Will run 115 specs

Running in parallel across 7 nodes

^CCanceling tests
2023/04/14 13:17:09 could not execute "test": {Error:<nil> StatusCode:137}
➜  kiln git:(feat/test-verbose-flag) ✗ ./kiln test -tp ~/workspace/tas/ist -v
Info: 13:17:18.442288 Checking for the latest ops-manager image...
Info: 13:17:18.442335 Building / restoring cached docker image
Info: 13:17:18.442338 This may take several minutes during updates to Ops Manager or the first run...
Info: 13:17:20.149812 Mounting /Users/nick/workspace/tas and testing ist
Info: 13:17:20.150036 Running: cd /tas/ist/test/manifest && PRODUCT=ist RENDERER=ops-manifest ginkgo -r -p -slowSpecThreshold 15
Running Suite: Manifest Generation Suite
========================================
Random Seed: 1681503440
Will run 115 specs

Running in parallel across 7 nodes

^CCanceling tests
```

crhntr:
I really like the documentation changes. They are helpful.

Also we should be thinking about improving logging. I like the changes in the logs.

I like how you wrote a small and purpose specific logger. That feels very "Go". However, Go 1.21 will likely have a structured logging library in the standard library. We can use that earlier by importing the experimental version here https://pkg.go.dev/golang.org/x/exp/slog. It might be nice to start getting feedback on it early to see if it would be a good library to use elsewhere.

crhntr (approved):
Looks good. Consider using the library I mentioned in the comment. If you prefer your implementation, I think that decision would also make sense.

notrepo05:
I'll start working on a follow up PR with the latest Go version, thanks for pointing me to the new logging library. I think this change could affect more than just test? 

For the sake of docs and quicker manifest testing I'm going to merge this for now.

PR #379, opened 2023-04-14
remove proofing.ExtractTileMetadataFile

crhntr:
This does the same thing and has tests: https://pkg.go.dev/github.com/pivotal-cf/kiln@v0.80.0/pkg/tile#ReadMetadataFromFile

Initially, I wanted to add test coverage; however, I noticed that I could just use the well tested implementation from the tile package. It has the same semantics as the original implementation. Then I thought, maybe we should just deprecate this function and recommend using the one from tile.

I also added some linter warning "fixes" in this commit.

### DIFF
--- a/pkg/planitest/manifest.go
+++ b/pkg/planitest/manifest.go
@@ -4,12 +4,12 @@ import (
 	"fmt"
 
 	"github.com/cppforlife/go-patch/patch"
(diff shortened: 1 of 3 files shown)


notrepo05 (approved):
Removes redundant code. Looks great!

PR #380, opened 2023-04-18
fix: Remove additional / from trainstat api endpoint

pabloarodas:


crhntr (approved):
Looks good :)

I'm not sure what is clearer fmt.Sprintf or just concatenating the strings with a +. I see this version more often and don't have a strong opinion.

pabloarodas:
> Looks good :)
> 
> I'm not sure what is clearer fmt.Sprintf or just concatenating the strings with a +. I see this version more often and don't have a strong opinion.

Thanks @crhntr , I went with the `fmt.Sprintf` since I saw that pattern being used in several places in the code. Both of them are clear and like you, I don't have a strong opinion, so I'll leave it like this.

ram-pivot (approved):
Looks good to me as well.

PR #383, opened 2023-05-01
feat: kiln publish attaches all matching license files to release

Spimtav:
Co-authored by: claire tinati <ctinati@vmware.com>


internal/commands/publish.go
```diff
@@ -374,35 +384,44 @@ func endOfSupportFor(publishDate time.Time) string {
...
 		}
 	}
-	return "", nil
+	return attachedLicenseFileNames, nil
 }
 
-func findMatchingOSL(productFiles []pivnet.ProductFile, version *releaseVersion) (pivnet.ProductFile, bool) {
+func findMatchingLicenseFiles(productFiles []pivnet.ProductFile, version *releaseVersion) ([]pivnet.ProductFile, bool) {
+	var matchingLicenseFiles []pivnet.ProductFile
+	foundLicenses := false
```
crhntr, line 416: This variable is not needed it is effectively the same as checking `len(matchingLicenseFiles)`. I don't think second return values gives us anything.

internal/commands/publish.go
```diff
@@ -374,35 +384,44 @@ func endOfSupportFor(publishDate time.Time) string {
...
-		err = p.PivnetProductFilesService.AddToRelease(slug, releaseID, productFile.ID)
-		if err == nil {
-			return productFile.Name, nil
-		} else {
-			return "", err
+		for _, licenseFile := range licenseFiles {
+			err = p.PivnetProductFilesService.AddToRelease(slug, releaseID, licenseFile.ID)
+
+			if err == nil {
+				attachedLicenseFileNames = append(attachedLicenseFileNames, licenseFile.Name)
```
crhntr, line 405: The control flow block here can be written without an ["else"](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow).

```
for _, licenseFile := range licenseFiles {
  if err := p.PivnetProductFilesService.AddToRelease(slug, releaseID, licenseFile.ID); err != nil {
    return attachedLicenseFileNames, err
  }
  attachedLicenseFileNames = append(attachedLicenseFileNames, licenseFile.Name)
}

File: internal/commands/publish.go (hunk 1 above)
Comment (crhntr) [reply internal/commands/publish.go:416]: Instead of matching license files. It think it might be clearer and less error prone if we list the product file IDs in the Kilnfile.

**Review by crhntr — CHANGES_REQUESTED** [review]
Nice PR. It fixes the problem precisely.

Please fix the control flow to pull the happy path to the left. After this change, I would be happy to merge.

An optional bigger change that might simplify this code would be to add a list of file IDs to the Kilnfile and just parse and use those instead of iterating through files and doing version string comparison.

File: internal/commands/publish.go (hunk 2 above)
Comment (Spimtav) [reply internal/commands/publish.go:405]: Noted and changed!  i've never seen that "call function, assign output, test output using semicolon, and do all this on the same line as the if-conditional" pattern before, but looking over the rest of the code and also some go idiom paged it seems like it's a common pattern?  i'm happy to continue using this style since it's convention!

Comment (Spimtav) [reply internal/commands/publish.go:405]: One additional note while i'm here - Danny and i decided to refactor the function to have an early return on the type being not GA, which un-nests a big block of code and might be a bit easier to follow.  Curious for your thoughts there!  It's semantically equivalent, but we both prefer early returns and thought it was worthwhile to do for readability.

File: internal/commands/publish.go (hunk 1 above)
Comment (Spimtav) [reply internal/commands/publish.go:416]: Did you one better - we don't even need a bool return, so i just made the caller of findMatchingLicenseFiles do the length checking instead!

**Comment by Spimtav** [comment]
Also worth noting potentially: Danny suggested that the tests can be made more robust by making the "errbuffer.NotTo.Contain(something)" lines shorter and less precise so they can catch a wider variety of changes.  We've done that in this commit as well!

File: internal/commands/publish.go (hunk 2 above)
Comment (crhntr) [reply internal/commands/publish.go]: I am also a fan of early returns. Sounds good! I'll take a look at the changed code.

File: internal/commands/publish.go [hunk 3]
```diff
@@ -374,35 +384,41 @@ func endOfSupportFor(publishDate time.Time) string {
...
-		}
+func (p Publish) attachLicenseFiles(slug string, releaseID int, version *releaseVersion) ([]string, error) {
+	var attachedLicenseFileNames []string
 
-		productFile, found := findMatchingOSL(productFiles, version)
-		if !found {
-			return "", errors.New("required license file doesn't exist on Pivnet")
-		}
+	if !version.IsGA() {
+		return []string{}, nil
```
crhntr, line 391: you can `return nil, nil`. Slices are fancy pointers so this does more than what's needed. It fine to do this but it adds an allocation.

crhntr, line 391: you can `return nil, nil`. Slices are fancy pointers so this does more than what's needed. It not a big deal in this case to return `[]string{}` but it adds an allocation and is more code than needed.

[See this comment for more details](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

crhntr (approved):

internal/commands/publish.go
Spimtav: Awesome, thanks for the context on this!  i've made that change and pushed it up, i'm not sure if it'll retrigger the code review, but if not, i'll just merge it in once the checks pass.

Thanks again for all the review on this PR, Christopher!

PR #394, opened 2023-05-26
fix acceptance tests after semver/v3 bump

crhntr:
The acceptance tests failure was caused by a change introduced in Masterminds/semver/v3. The comparison semantics for major 0 changed. The problem is summarized in this playground: https://go.dev/play/p/t5beNbYVR9K. I changed the zero value of the cargo.ComponentSpec version field from ">0" to ">=0" go get the same behavior as before.

The first commit in this branch fixes the issue. I found some other issues with the acceptance tests that caused failures on MacOS. I fixed those and refactored to make build failures easier to understand.

### DIFF
--- a/internal/acceptance/workflows/acceptance_test.go
+++ b/internal/acceptance/workflows/acceptance_test.go
@@ -13,8 +13,10 @@ package workflows
 
 import (
 	"context"
+	"golang.org/x/exp/slices"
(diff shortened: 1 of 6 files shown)


jaristiz (approved):
The changes seems OK, I like the refactors made. 👍 from me.

PR #395, opened 2023-06-01
Remove credentials requirement on Artifactory (release source) read operations

crhntr:
Story: 185151052

PR #396, opened 2023-06-01
fix: dont authenticate GETs to artifactory

notrepo05:
- remove basic auth from find and match release in artifactory_release_source

[https://www.pivotaltracker.com/n/projects/2535033/stories/185151052]


notrepo05:
@crhntr @pabloarodas and I believe this fixed the issue. But I vaguely recall a discussion about non-authenticated endpoints not returning all of the releases.. if so, this will just introduce a new bug. We couldn't find any issues though..

crhntr:
This is fixing the same thing: https://github.com/pivotal-cf/kiln/pull/395

pabloarodas (approved):
LGTM

PR #397, opened 2023-06-06
only release v0 releases

crhntr:
We need to stabilize the flags, commands, **and package api** before we ship 1.0.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,9 @@ name: release
 
 on:
   push:
-    tags: ['v*']
+    tags: ['v0*']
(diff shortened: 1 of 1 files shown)


crhntr:
We are too late. It looks like we need to push a v2 release version to make any breaking changes. I was really hoping to avoid this! It will require a bunch of work.

If we can pull this release, we should look into doing that.

Screenshot 2023-06-06 at 4 44 21 PM

notrepo05 (approved):

notrepo05:
Brew, dockerhub, and the go package cache should all default to the latest 0.x / retract the major 1 tags.

PR #401, opened 2023-06-08
support sts assume role with s3 release sources

abg:
Adds an optional "aws_role_arn" property for a release source that can assume a role in an aws account.

[#185364456](https://www.pivotaltracker.com/story/show/185364456)


crhntr:
I cherry-picked the commit from this branch and refactored it a little bit. Thank you for the contribution!

Issue #403, opened 2023-06-15
"github.com/pivotal-cf/pkg/tile".ReadMetadataFromFile does not read metadata/some-product.yml


crhntr:
### Describe the bug

Not all tile builder write the metadata to "metadata/metadata.yml" and OpsManager supports other names.

### Reproduction steps

1. checkout some tile-generator tile (for example credhub-service-broker)
2. tile build
3. try to read the metadata from the output tile using `tile.ReadMetadataFromFile`

### Expected behavior

```go
	zipFile, _ := zip.OpenReader(tileGeneratorOutputFileName)
	matches, _ := fs.Glob(zipFile, "metadata/*.yml")
	if len(matches) == 0 {
		// ensure a match is found
	}
	metadataBuffer, _ := fs.ReadFile(zipFile, matches[0])
```

### Additional context

We should dig into what the ops manager behavior would be if there are multiple files (how does it pick the "first" one)

If Christopher quits and opens a coffee shop, @Tallicia has context.

PR #404, opened 2023-06-16
feat: parse non-standard metadata filenames

crhntr:
tile-generator creates metadata filenames that are not "metadata/metadata.yml" with this change you can parse those tiles with the tile.ReadMetadata* functions

fixes: #403

### DIFF
--- a/pkg/tile/metadata.go
+++ b/pkg/tile/metadata.go
@@ -30,7 +30,15 @@ func ReadMetadataFromZip(ra io.ReaderAt, zipFileSize int64) ([]byte, error) {
 }
 
 func ReadMetadataFromFS(dir fs.FS) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


staylor14:
@crhntr what are we supposed to do with multiple metadata files?

staylor14 (commented):
I believe we may need to know how to handle multiple files if OpsMan does.

crhntr:
That is scary... I did not know it did that.

staylor14 (commented):
Turns out, OpsMan globs like this does: the first *.yml file in the metadata directory is assumed to be the one and only file.  [Slack convo](https://vmware.slack.com/archives/C01EBT9F3NF/p1687907842741389)

staylor14 (approved):

PR #328, opened 2022-08-22
[DO NOT MERGE] Kiln 1.0.0 -- deprecations and refactors

crhntr:
# High-level Changes
- [x] [Rename commands and flags to match our design document](https://github.com/pivotal-cf/kiln/pull/328/commits/0ca38ea123ea864e3f17618f7dff849fd6afcd9e)
- [x] Remove OM_PRIVATE_KEY (use private-key query param from BOSH_ALL_PROXY instead)
- [x] Remove publish command [950acde](https://github.com/pivotal-cf/kiln/pull/328/commits/950acdef85e24c11c14d984234c77c142dc1925c)
- [x] hide WIP packages in internal (proofing and manifest)
- [x] standardize test imports (use `.` instead of `Ω` for Gomega)
- [x] help output is grouped by usage
- [x] refactor release sources
  - [x] make release sources load from yaml directly (remove factory type) 
  - [x] remove MultipleReleaseSource interface
  - [ ] merge component and cargo packages (or at least move component to a sub directory of cargo)
- [ ] add a version number to the kilnfile
- [ ] improve release-source credential loading maybe from environment variables that match CLIs (have fallbacks for when the a Kilnfile does not specify the values as variables)
- [ ] update-stemcell respects Kilnfile stemcell_criteria version constraint

PR #405, opened 2023-06-28
Make where we work with BOSH Release tarballs explicit

crhntr:
## TL;DR

- `component.Spec` and `cargo.ComponentSpec` -> `cargo.BOSHReleaseTarballSpecification`
- `component.Lock` and `cargo.ComponentLock` -> `cargo.BOSHReleaseTarballLock`
- `cargo.ReleaseSourceTypeBOSHIO` -> `cargo.BOSHReleaseTarballSourceTypeBOSHIO`
- `cargo.ReleaseSourceTypeS3` -> `cargo.BOSHReleaseTarballSourceTypeS3`
- `cargo.ReleaseSourceTypeGithub` -> `cargo.BOSHReleaseTarballSourceTypeGithub`
- `cargo.ReleaseSourceTypeArtifactory` -> `cargo.BOSHReleaseTarballSourceTypeArtifactory`
- *a few other function and method names were changed in a similar way to align with the above rename operations.*

## Discussion

This refactor is intended to make it easier to think about and code around the BOSH releases tarballs we package.

- In 96032c82cec94316a3784b5960ded477afd2a761, I merge `component.Lock` and `cargo.ComponentLock`. They have [literally been the same thing for a while](https://github.com/pivotal-cf/kiln/blob/57a648139d891c36ea43a113d22545bb3efda0f3/internal/component/component.go#L9-L22); this commit removes the indirection.
  - `component.Local` and `component.Exported` no longer embed the `cargo.ComponentLock` but have it as a non-embedded field. This makes refactors easier and makes it clear that those types should not inherit methods from the Lock. For the most part, this change is reversible in a non-breaking way.
- In a67fdfd0fd92cf5db8738b42d644f0efea7092a2, I rename "Component{Lock,Spec} to BOSHRelease{Lock,Spec}. We may still want to refer to BOSH Releases as tile components in speech and docs but the code reads better when it is less ambiguous about what is is operating on.
- In f0f1656d3c38e4ad1f99bf15f8b5245b63c38bf1, I add the Tarball part to the BOSHRelease names.
- In 797771017b28b27e6f8775ef77180e6031a76cd9, I renamed other public identifiers and methods to stay consistent with the Tarball Lock and Spec types.


crhntr:
[Prior art](https://github.com/pivotal-cf/kiln/pull/328/commits/4fc5bf46a2e927d3178132d55e895d13f1cd5676)

dtimm (approved):
Excellent refactor. LGTM.

PR #406, opened 2023-06-28
Add feature to specify which BOSH Release Tarball versions are locked and how BOSH Release Tarball versions are unlocked

crhntr:
## New Behavior

- **`kiln glaze`**: pins all releases to exact version in current Kilnfile.lock

  The release in the Kilnfile is skipped if it has the field value `always_float: true`

- **`kiln glaze --undo`**: "unpins" by pinning component version according to glaze_behaivor

  The release in the Kilnfile gets it's version constraint set based on the value of `maintenance_version_bump_policy`. 

  The permitted values are:

  - "LockNone"
  - "LockMajor" (this is also the default for when the value is the empty string or is not set)
  - "LockMinor"
  - "LockPatch"

  The stemcell version constraint is not changed by this command.

## Code Review Notes

The business logic is in the cargo package the command code does not exercise the functionality.

## `github.com/pivotal-cf/kiln/internal/commands`

The Glaze command tests do not test functionality; they exercise the flag parsing and pass values to faked out functions.

[The rarely used syntax to convert a method to an function that can be passed/assigned has a simple example here to make understanding it in the PR code easier.](https://go.dev/play/p/C_tv6ccNsKj)

### Error Behavior
- When the version of the component in the Kilnfile.lock is not valid, the error is just returned.
- When the input `string`/`[]byte` or `cargo.DeGlazeBehavior` value is unknown, an error with message "unknown behavior" is returned.
- When parsing fails, the error is just returned.

### Notes

- `cargo.DeGlazeBehavior.String` returns an empty string when the value of `DeGlazeBehavior` is unknown.
- I add a dependency on a ["bijection"](github.com/crhntr/bijection) data structure that I wrote a while back as an exploration of generics. It is not required and the idiomatic way would be to use ["Stringer"](https://pkg.go.dev/github.com/golang/tools/cmd/stringer). This is internal behavior and can be refactored in the future.
- Both `cargo.Kilnfile.Glaze` and `cargo.Kilnfile.DeGlaze` return an error with a message like `"release with name "banana" not found in Kilnfile.lock: not found"` when the Lock does not contain a spec found in the Kilnfile.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,6 +63,7 @@ require (
 	github.com/containerd/containerd v1.6.18 // indirect
 	github.com/containerd/typeurl v1.0.2 // indirect
 	github.com/cppforlife/go-semi-semantic v0.0.0-20160921010311-576b6af77ae4 // indirect
+	github.com/crhntr/bijection v0.0.0-20230628013949-46b5c800bc70 // indirect
 	github.com/cucumber/gherkin-go/v19 v19.0.3 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/docker/distribution v2.8.1+incompatible // indirect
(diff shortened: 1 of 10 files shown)


notrepo05 (approved):
Already reviewed. Quickly reviewed, LGTM

PR #408, opened 2023-07-03
component: consistently handle not found errors

abg:
When a github-release source is not found, kiln aborts early without trying additional release sources.

This change ensures the github release source reports "ErrNotFound" so that subsequent release sources can be tried.

Bumped go.mod -go=1.20 to pull in "errors.Join".  kiln is already built with Go 1.20 so this seems reasonable.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/pivotal-cf/kiln
 
-go 1.18
+go 1.20
 
 require (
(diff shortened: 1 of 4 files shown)


crhntr:
I am not sure why checkout was failing. I created a new PR here: https://github.com/pivotal-cf/kiln/pull/430 thinking it might be something with repo permissions in actions.

PR #410, opened 2023-07-05
add BOSH Release tarball reading utilities

crhntr:
This adds some utilities for reading BOSH release manifests (release.MF) from tarballs.

I refactor the builder and acceptance tests to use this to show that this makes sense as a place for these utility functions. My main intent for adding them was to use them in the kiln buildpacks.


jaristiz (approved):
Like the refactor to the `release` package, found it confusing when using `github.com/kiln/pkg/tile` package in other projects.

crhntr:
I like it too. Maybe tilerelease? Unfortunately, "release" is such an overloaded term.

dtimm:
Why add this to `kiln` specifically? I think it would make more sense as a separate repository that is used in `kiln` rather than included wholesale. Since it isn't used (yet), could the PR wait until the functionality is needed?

crhntr:
We do use this during bake.

PR #413, opened 2023-07-06
feat: allow TanzuNet slug to be set in the Kilnfile stemcell criteria

crhntr:
For a while we have maintained a switch statement that mapped stemcell slugs like "ubuntu-jammy/1.2" to product slugs like "stemcells-ubuntu-jammy". This was annoying but not that big of an issue because it we do it every year or so. Instead of continuing this pattern, lets add a field to the Kilnfile so the CLI does not need to change when a new stemcell is released.

This change helps me with my buildpack work because, I want to be able to compile releases against stemcells and I want the particular stemcell to be specified in the tile source. I don't want to maintain the same switch statement in both Kiln and in the buildpack. I intend to import these functions into the buildpack.

### DIFF
--- a/internal/commands/find_stemcell_version.go
+++ b/internal/commands/find_stemcell_version.go
@@ -52,15 +52,9 @@ func (cmd FindStemcellVersion) Execute(args []string) error {
 		return err
 	}
 
-	productSlug := ""
-
-	// Get Stemcell OS and major from Kilnfile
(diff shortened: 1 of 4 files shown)


crhntr:
@jaristiz I vaguely remember talking with you about this change a long time ago.

dtimm (approved):
LGTM.

PR #415, opened 2023-07-11
Kiln cannot parse bosh variables when given as top level array.

### LINKED ISSUE #369: Kiln unable to parse variables block
### Describe the bug

The latest commit of kiln is no longer able to bake the RabbitMQ tile and instead fails with the following error:
```
could not execute "bake": failed to parse template variables: unable to YAML parse "variables/erlang_cookie.yml": yaml: unmarshal errors:
  line 2: cannot unmarshal !!seq into map[string]interface {}
```

The variables file in question has the following content:
```
- name: erlang_cookie
  type: password
  options:
    length: 22
```
and is referenced from a variables block in the base.yml as
```
variables:
- $( bosh_variable "erlang_cookie" )
```

The variable is referenced elsewhere [using triple parenthesis notation, per the tile developer guide](https://docs.vmware.com/en/Tile-Developer-Guide/3.0/tile-dev-guide/property-template-references.html#variables-23), to create a manifest snippet using triple parenthesis notation as
```
named_manifest:
- name: erlang_cookie
  manifest: "(((erlang_cookie)))"
```
This manifest snippet is then consumed as `parsed_manifest(erlang_cookie)`

### Reproduction steps

Attempt to bake RabbitMQ tile. Observe failure parsing YAML snippet due to it being an array rather than a hash.

### Expected behavior

bake should parse the manifest snippet.

### Additional context

The tile parsed successfully using v77.3, so some change between v0.77.3 and v0.78.1 has caused this parsing failure.

MirahImage:
Resolves #369

### DIFF
--- a/example-tile/bosh_variables/some-ca-certificate.yml
+++ b/example-tile/bosh_variables/some-ca-certificate.yml
@@ -1,5 +1,5 @@
-- name: some-ca-certificate
-  type: certificate
-  options:
-    common_name: 'ca.example.com'
-    is_ca: true
+name: some-ca-certificate
+type: certificate
+options:
+  common_name: 'ca.example.com'
+  is_ca: true
--- a/example-tile/bosh_variables/some-secret.yml
+++ b/example-tile/bosh_variables/some-secret.yml
@@ -1,2 +1,2 @@
-- name: some-secret
-  type: password
+name: some-secret
+type: password


crhntr:
I tried to re-open the PR so it runs again. I am not sure why the "missing token" error is occurring.

crhntr (approved):

PR #416, opened 2023-07-11
add S3 aws role arn

crhntr:
I created this PR based on #401 to do some refactors before merging the changes @abg made.


ram-pivot (approved):
Changes look good to me. I was also manually able to test AWS bucket access using assume role

PR #417, opened 2023-07-11
ci: add action to create new kiln releases when a newer Go is available

crhntr:
This automates a recently process that @marroyo and I went through.

We found that the latest release did not use the latest Go.
We created a tag with the same revision as the previous release.
We watched the release Github action do it's thing.


pabloarodas:
@crhntr could you please give me some context on why create a new release built with the latest go, without bumping anything in the go.mod file?
Just want to make sure I understand the reason behind this new action

PR #419, opened 2023-07-14
refactor: replace the empty interface with any

crhntr:
The "any" keyword was added in 1.18. Let's use it.


crhntr:
already in a different PR

PR #422, opened 2023-07-27
make find-release-version output match Kilnfile.lock

nouseforaname:
We from cryo just started using kiln for our tile builds. We have quite a few tiles and mostly want to bump versions in a tile to latest since we don't ship many releases in a tile.
I found the output of the find-release-version command somewhat confusing because it doesn't really match the structure of the Kilnfile.lock so I figured I'd try to PR this in.

What does this do? It drops the `releaseVersionOutput` that wasn't used anywhere else and just uses `BOSHReleaseTarballLock` to generate the output.

It adds the option to output yaml (in case you want to use the output and put it into the lockfile..)

it also updates the `BOSHReleaseTarballLock` with json tags so marshalling produces the expected output.

### DIFF
--- a/internal/commands/find_release_version.go
+++ b/internal/commands/find_release_version.go
@@ -7,6 +7,7 @@ import (
 	"log"
 
 	"github.com/pivotal-cf/jhanda"
+	"gopkg.in/yaml.v3"
 
 	"github.com/pivotal-cf/kiln/internal/commands/flags"
 	"github.com/pivotal-cf/kiln/pkg/cargo"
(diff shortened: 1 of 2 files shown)


pabloarodas:
@nouseforaname I opened [a new PR](https://github.com/pivotal-cf/kiln/pull/436) to run our test CI since it doesn't run on forks. I also fixed JSON output and added tests for it. Closing this PR on behalf of the new one.

nouseforaname:
Thanks!

pabloarodas:
@nouseforaname Sorry about this but after checking with the team, we found that the current output of the `kiln find-release-version` cannot be changed since it is being used by other tools in our TAS CI. Having an additional yaml output but with different fields wouldn't affect those tools but would make the command not have a consistent output for the two formats which is not desirable for us.

We would like to understand your use-case for the yaml output so we can find an alternative to what you need to do. Could you please explain why would you need the `kiln find-release-version` to match the Kilnfile.lock format?

PR #424, opened 2023-07-31
add compile-bosh-releases-command

crhntr:
This change adds a `kiln compile-bosh-release-tarballs` command.

```
$ kiln compile-bosh-release-tarballs --help
kiln compile-bosh-release-tarballs
upload bosh release tarballs to a bosh director, compile the packages against the stemcell from the stemcell lock, and download the compiled tarballs

Usage: kiln [options] compile-bosh-release-tarballs [<args>]
  --help, -h     bool  prints this usage information (default: false)
  --version, -v  bool  prints the kiln release version (default: false)

Command Arguments:
  --kilnfile            string  path to Kilnfile (default: Kilnfile)
  --releases-directory  string  path to a directory containing release tarballs (default: releases)
  BOSH_ALL_PROXY        string
  BOSH_CA_CERT          string
  BOSH_CLIENT           string
  BOSH_CLIENT_SECRET    string
  BOSH_ENVIRONMENT      string
  BOSH_ENV_NAME         string
```
Note the only way to set the BOSH variables is via the environment. These variables match the ones set by smith and BBL.

## TODO
- [x] design the command name and flags
- [x] write an implementation that takes local tarball files and writes them locally
- [ ] add flag to specify which release source to upload to. The flag should be the similar flag in `cache-compiled-releases` and should be named `--upload-target-id`
- [x] update the Kilnfile.lock after the download/export succeeds


crhntr:
This is not longer how we want to do this. We may wrap an image with https://github.com/cloudfoundry/bosh-agent/pull/315 in the future but for now this PR should be closed.

PR #426, opened 2023-08-02
feat: change --skip-fetch-directories to boolean

notrepo05:
skip all or nothing instead of passing specific directories

Authored-by: Nick Rohn <nrohn@vmware.com>

### DIFF
--- a/internal/commands/bake.go
+++ b/internal/commands/bake.go
@@ -176,7 +176,7 @@ type Bake struct {
 		Sha256                   bool     `            long:"sha256"                                                description:"calculates a SHA256 checksum of the output file"`
 		StubReleases             bool     `short:"sr"  long:"stub-releases"                                         description:"skips importing release tarballs into the tile"`
 		Version                  string   `short:"v"   long:"version"                                               description:"version of the tile"`
-		SkipFetchReleases        bool     `short:"sfr" long:"skip-fetch-directories"                                description:"skips the automatic release fetch for all release directories"`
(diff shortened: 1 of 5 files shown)


notrepo05:
After tests pass I'd request that this is manually tested before merging as well. Thank you

internal/commands/bake.go
```diff
@@ -176,7 +176,7 @@ type Bake struct {
 		Sha256                   bool     `            long:"sha256"                                                description:"calculates a SHA256 checksum of the output file"`
 		StubReleases             bool     `short:"sr"  long:"stub-releases"                                         description:"skips importing release tarballs into the tile"`
 		Version                  string   `short:"v"   long:"version"                                               description:"version of the tile"`
-		SkipFetchReleases        []string `short:"sfr" long:"skip-fetch-directories"        description:"skips the automatic release fetch the specified release directories"`
+		SkipFetchReleases        bool     `short:"sfr" long:"skip-fetch-directories"                                description:"skips the automatic release fetch for all release directories"`
```
crhntr, line 179: Would you please introduce a new flag "--skip-fetch" and "--skip-fetch-directories" as an alias to it. https://github.com/pivotal-cf/kiln/tree/feature-export-bosh-release-download-methods

crhntr (changes requested):

internal/commands/bake.go
notrepo05: Done. I'd suggest testing this locally before merging

crhntr (approved):

PR #434, opened 2023-08-17
add a utility function to the tile package for reading metadata from a remote tile file

crhntr:
This will be helpful for making kiln test tile revisions against recently released tiles.

I intend to use this on a few upcoming features but am adding it now because this function is a helpful utility that TAS Operability may get value from.

I intentionally did not use the pivnet package because we use various versions of that module in this package. Once https://github.com/pivotal-cf/kiln/pull/318 is merged, we may want to re-evaluate this API. As it is, this implementation might work with Artifactory.


crhntr:
I accidentally used the wrong base branch.

PR #439, opened 2023-09-05
refactor(commands): remove unused vault function

crhntr:
I don't see this used anywhere. It adds a few vault dependencies. Hashicorp recently switched to BSL license. I don't know if Kiln would be in violation, but to keep it simple, we don't use this function so let's remove it.


pabloarodas:
Thanks for doing this Chris, looks good to me.

pabloarodas (approved):

PR #441, opened 2023-09-08
Don't always add default values

crhntr:
Kiln currently will behave strangely when an options structure has multiple lists of string fields. Jhanda adds default flag values even when flags are passed. Kiln trims these additional defaults off when a user passes any flags that match that field. There was an error in the logic where it would return early and not trim defaults on fields after the first configured list.

```go
type Command struct {
  Options struct{
    ListA []string `short:"a" default:"apple"`
    ListB []string `short:"b" default:"banana"`
  }
}
```

Given the above struct,
if you ran `cmd -a=apricot -b=bean`,
then you'd experience `cmd -a=apricot -b=banana -b=bean`

I finally found and fixed a bug here a5bd61fe79d6734ff1b0f95db8c4ffc4145b4b20. The rest of the commits were small refactors/improvements that I found along the way.

The commits headers were written to be read. I decided not to squash these changes to make review easier. The hunt for the bug was chaotic and I feel this is safe to ship in a patch release.


pabloarodas (approved):
Looks good to me. Sorry for the delay in reviewing it, It took me long enough to understand the changes.

PR #442, opened 2023-09-08
refactor(acceptance/bake): use tile pkg and isolate tile source

crhntr:
this is experiment towards simplifying tests to just use temporary test directories and the tile package


crhntr:
Not really needed. We can do the same stuff in another PR.

PR #443, opened 2023-09-14
Update README.md

dalvarado:
remove specific version that is checked out in the "build from source" section


pabloarodas (approved):

pabloarodas:
Thanks for finding and fixing this.

PR #444, opened 2023-09-15
add additional logs when removing releases

pnikonowicz:
when a release is removed, it is difficult to understand the reason why. this adds some helpful messages to better communicate if a bad sha or a bad version is the reason for deleting a mismatched release

### DIFF
--- a/internal/commands/fetch.go
+++ b/internal/commands/fetch.go
@@ -164,9 +164,9 @@ nextRelease:
 				missing = append(missing[:j], missing[j+1:]...)
 				continue nextRelease
 			} else if rel.Lock.Name == lock.Name && rel.Lock.Version == lock.Version {
-				fmt.Printf("release: [ %s ] sha mismatch: [ %s ]\n", lock.Name, rel.Lock.SHA1)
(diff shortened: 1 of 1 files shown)


internal/commands/fetch.go
```diff
@@ -163,6 +163,10 @@ nextRelease:
 				intersection = append(intersection, rel)
 				missing = append(missing[:j], missing[j+1:]...)
 				continue nextRelease
+			} else if rel.Lock.Name == lock.Name && rel.Lock.Version == lock.Version {
+				fmt.Printf("release: [ %s ] sha mismatch: [ %s ]\n", lock.Name, rel.Lock.SHA1)
```
pabloarodas, line 167: Thanks for doing this @pnikonowicz, could you please just add the word local to both messages so the message is clear that we are printing the information for local releases, something like: `fmt.Printf("Local release: [ %s ] sha mismatch: [ %s ]\n", lock.Name, rel.Lock.SHA1)`

pabloarodas (changes requested):
Could you please make that small change, besides that this change makes sense and it looks good to me.

pabloarodas (approved):

pabloarodas (approved):

pabloarodas:
Closing this on behalf of #445, since our CI does not run well on fork PRs.

PR #446, opened 2023-10-04
Kiln bake logs confusing error when the tarball is empty (this happens sometimes when fetch fails)

crhntr:
A confusing error happens sometimes when fetch fails. This change makes the error mesage clearer.

Instead of the error
> could not execute "bake": failed to parse releases: EOF

the user should see

> could not execute "bake": failed to parse releases: BOSH release tarball some-release-1.2.3.tgz is an empty file

We should also make sure fetch is cleaning up after itself on failure.

### DIFF
--- a/pkg/cargo/bosh_release.go
+++ b/pkg/cargo/bosh_release.go
@@ -11,6 +11,7 @@ import (
 	"io/fs"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


pabloarodas (approved):
Looks good to me. Thanks for doing this.

PR #458, opened 2024-01-05
Update README.md

markstokan:
Kiln currently doesn't support bash process substitution so we explicitly write credentials to filesystem.


pvaramballypivot (approved):

markstokan:
unrelated unit test failure.  merging since this is only a README change.

PR #462, opened 2024-01-12
make acceptance tests less error prone when changes in hello-tile occur

crhntr:
This change makes each test use a temporary tile source directory and the source is copied from simple directories. Git is initialized for each test. Most changes to crhntr/hello-tile source files will no longer break kiln tests.

In both of these cases, the tests will fail with a clear error message.
- Changes to crhntr/hello-tile milestones and issues might break kiln tests.
- Removals of old crhntr/hello-release GitHub releases will break kiln tests.


crhntr:
the test failed and I have a more stable fix for that error here: https://github.com/pivotal-cf/kiln/pull/462

PR #467, opened 2024-01-16
Exclude buildpacks and stacks while glazing

pvaramballypivot:


crhntr:
This functionality is already supported by the "AlwaysFloat" field.

See: https://pkg.go.dev/github.com/pivotal-cf/kiln@v0.89.0/pkg/cargo#BOSHReleaseTarballSpecification

pvaramballypivot:
@crhntr I look through the AlwaysFloat. It wasn't clear that we could use it per release. Il take a look at it. Thanks.

PR #469, opened 2024-01-17
Remove extra property addition to Kilnfile

pvaramballypivot:
Print releases not glazed

Rename float_always to skip_glaze

### DIFF
--- a/pkg/cargo/kilnfile.go
+++ b/pkg/cargo/kilnfile.go
@@ -36,8 +36,10 @@ func (kf *Kilnfile) BOSHReleaseTarballSpecification(name string) (BOSHReleaseTar
 
 func (kf *Kilnfile) Glaze(kl KilnfileLock) error {
 	kf.Stemcell.Version = kl.Stemcell.Version
+	var SkipGlazeReleases []string
 	for index, spec := range kf.Releases {
 		if spec.FloatAlways {
+			SkipGlazeReleases = append(SkipGlazeReleases, spec.Name)
 			continue
 		}
 		lock, err := kl.FindBOSHReleaseWithName(spec.Name)
@@ -46,6 +48,15 @@ func (kf *Kilnfile) Glaze(kl KilnfileLock) error {
 		}
 		kf.Releases[index].Version = lock.Version
 	}
+	if len(SkipGlazeReleases) > 0 {
+		fmt.Println("Skipping glaze for the following releases:")
+		for _, release := range SkipGlazeReleases {
+			fmt.Println(release)
+		}
+	} else {
+		fmt.Println("All releases in the Kilnfile are glazed")
+	}
+
 	return nil
 }
 
@@ -111,11 +122,11 @@ type BOSHReleaseTarballSpecification struct {
 	GitHubRepository string `yaml:"github_repository,omitempty"`
 
 	// DeGlazeBehavior changes how version filed changes when de-glaze is run.
-	DeGlazeBehavior DeGlazeBehavior `yaml:"maintenance_version_bump_policy"`
+	DeGlazeBehavior DeGlazeBehavior `yaml:"maintenance_version_bump_policy,omitempty"`
 
 	// FloatAlways when does not override version constraint.
 	// It skips locking it during Kilnfile.Glaze.
-	FloatAlways bool `yaml:"float_always,omitempty"`
+	FloatAlways bool `yaml:"skip_glaze,omitempty"`
 
 	// TeamSlackChannel slack channel for team that maintains this bosh release
 	TeamSlackChannel string `yaml:"slack,omitempty"`


pkg/cargo/kilnfile.go
```diff
@@ -36,8 +36,10 @@ func (kf *Kilnfile) BOSHReleaseTarballSpecification(name string) (BOSHReleaseTar
 
 func (kf *Kilnfile) Glaze(kl KilnfileLock) error {
 	kf.Stemcell.Version = kl.Stemcell.Version
+	var SkipGlazeReleases []string
```
crhntr, line 39: please lowercase the first character in the identifier

pkg/cargo/kilnfile.go
```diff
@@ -111,11 +122,11 @@ type BOSHReleaseTarballSpecification struct {
 	GitHubRepository string `yaml:"github_repository,omitempty"`
 
 	// DeGlazeBehavior changes how version filed changes when de-glaze is run.
-	DeGlazeBehavior DeGlazeBehavior `yaml:"maintenance_version_bump_policy"`
+	DeGlazeBehavior DeGlazeBehavior `yaml:"maintenance_version_bump_policy,omitempty"`
 
 	// FloatAlways when does not override version constraint.
 	// It skips locking it during Kilnfile.Glaze.
-	FloatAlways bool `yaml:"float_always,omitempty"`
+	FloatAlways bool `yaml:"skip_glaze,omitempty"`
```
crhntr, line 129: I recall @staylor14 asking for the name "float_always". I don't recall the exact case. I believe it had something to do with describing the behavior in connection with the deglaze policy names.

My best argument for the field "float_always" is I always sing a few lines of Janelle Monáe's "Float" when I see the field "always_float" and it makes me happy.

pkg/cargo/kilnfile.go
```diff
@@ -36,8 +36,10 @@ func (kf *Kilnfile) BOSHReleaseTarballSpecification(name string) (BOSHReleaseTar
 
 func (kf *Kilnfile) Glaze(kl KilnfileLock) error {
 	kf.Stemcell.Version = kl.Stemcell.Version
+	var SkipGlazeReleases []string
 	for index, spec := range kf.Releases {
 		if spec.FloatAlways {
+			SkipGlazeReleases = append(SkipGlazeReleases, spec.Name)
```
crhntr, line 42: week opinion: consider logging the skip and maybe setting a boolean "hasAlwaysFloatReleases" instead of creating another list... I don't see any other use of the list.

crhntr (changes requested):

pkg/cargo/kilnfile.go
pvaramballypivot: "float_always" could confuse the component teams to think that it is equivalent to the version constraint "*" and they may end up adding this property to their releases.

pkg/cargo/kilnfile.go
pvaramballypivot: @crhntr Can you please elaborate? I am super new to Go

pkg/cargo/kilnfile.go
crhntr: Removing that source of confusion makes sense to me. I think I like the change now.

I'll reach out to @staylor14 to see if he has a strong opinion.

pkg/cargo/kilnfile.go
crhntr: Sure :) You seem to be adding elements to a list to them immediately log the lines out. Doing so adds an extra data structure (a simple slice) that you may not need. You could get the same result with logging in the `if spec.FloatAlways {}` block. If you really want to add the "all releases were glazed" log line, you could keep that functionality by assigning "skippedSomeValue = true" or something like that then conditionally logging, although I don't think that tells me anything useful that I don't already know from the other log lines.

pvaramballypivot: Ha yes. That makes total sense. When I am working on kiln code, I tend to get lost and don't think straight lol. Thank you.

pvaramballypivot: Ok at first I was wondering why I didn't think of this. But now I know. 
I want to display the message "Skipping glaze for the following releases:" only if it skips pinning releases.
And I won't know that until it processes all the releases in the Kilnfile. That's why I am not printing the skipped releases in the spec.FloatAlways {} block.

pkg/cargo/kilnfile.go
staylor14: So, "\*" is a **kind** of float.  `glaze` will pin that. "1.*" is **also** a float.  `glaze` will pin that.

In my head, a Component team is usually asking for some kind of float, so the directive that made sense to me in the face of a `glaze` command was `float_always`.

If it makes more sense to more people to put this in terms of the command (glaze) rather than the effect (floating), that's cool with me, BUT!

I do not like booleans that are named for negation (e.g. `do_not_do`), nor do I like having a difference between internal and external names (`FloatAlways` vs `skip_glaze`).  I propose we re-do both names: `AllowGlaze` and `allow_glaze`.  Note that this will require a flip to the actual logic in-play, but that's to be expected because the semantic meaning of the words has now flipped (`float_always=true` meant "never glaze", and now `allow_glaze=true` will mean the opposite).

How'z about them apples? Who likes the apples?

staylor14: Also, is the phrase "semantic meaning" redundant?  I think it's redundant. Yeah, it's redundant. Redundant.

pvaramballypivot:
Closing this as we decided to change the approach.

PR #481, opened 2024-03-06
Golden Path Changes - Part 2

crhntr:
**TL;DR: This change set includes a new command, minor changes to kiln bake, and some public changes to the Go packages.**

The theme for these changes is supporting the centralized build system for Tanzu Tiles. 

Tiles will now be (sha256) identical between kiln versions.

### Omit Kiln Version from Product Template

`kiln bake` no longer adds the kiln version to the product template. To keep track of which kiln version was used to bake a tile, authors should use the `--final` flag and commit the resulting bake record to the repository. I considered falling back to add the kiln_version when the --final flag was not used but can add that later if needed and want to propose the smallest change set.

### Add a `kiln re-bake` Command

The new command `kiln re-bake` takes a path to a bake record and builds a tile byte for byte like it was built before. I intend to merge this into the kiln bake command but the control flow in the bake command is not clear and this proposed new command keeps the change as small as possible. As bake is refactored, we can merge in the functionality with reduced risk.

### Add Tile Path to Bake Record

The Bake record now keeps track of a relative path from the repository root to the tile root (aka where the Kilnfile exists). This will make restructuring tile repos easier and reduce coupling between Golden Path configuration and tile repositories across revisions.

### Create a `bake` Package

This will be the destination for refactored baking code. This package currently contains the `bake.Record` which is a data-structure for checksums and metadata about a tile build. The bake record maps source revisions, kiln versions, and tile checksums.


internal/commands/rebake.go
```diff
@@ -0,0 +1,87 @@
...
+	if !record.IsEquivalent(newRecord, log.New(os.Stderr, "bake record diff: ", 0)) {
+		return fmt.Errorf("expected tile bake records to be equivilant")
+	}
+
+	return nil
+}
+
+func (cmd ReBake) Usage() jhanda.Usage {
+	return jhanda.Usage{
+		Description:      "re-bake (aka record bake) builds a tile from a bake record if the current HEAD is does not match the record the command will fail",
```
pabloarodas, line 85: Could you please clarify this Description, this part `if the current HEAD is does not match the record the command will fail` is not clear

pabloarodas:
We reviewed the PR with Riz and Preethi and we think the changes are ok. Derek gave us with some context about the Golden path that helped us understand better the reason behind these changes, however, one feedback we have for future PRs, is to try and separate feature work from refactors since those make the PRs too big and not so easy to review.

@crhntr please let us know when you finish doing changes in the PR and mark it as `Ready for Review` so we can go ahead and merge it.

crhntr:
I spent all day Friday trying to figure out why the acceptance test was sometimes failing on new commits in CI. I did not figure it out. I will pick it up when I get back.

pabloarodas (approved):

PR #482, opened 2024-03-07
Fixes bug when reading tile_name variable

ryanwittrup:
Our team recently experienced CI failures when trying to bake a tile, after passing in the [tas variables file](https://github.com/pivotal/tas/blob/main/tas/variables/srt.yml)

That file has the properties:
`tile_name: SRT`

The result was a strange error: 'tile_name value must be a string got value "SRT" with type string'

After investigating more, we found [this commit](https://github.com/pivotal-cf/kiln/commit/ce05d124b98f985d14c700afebfb8a9d0eb13e8d#diff-cf90d1c4858c9124c40badad400dbe3aafed9ad0c7a3c72b36e14532b6d6bf25R588) included a small typo in how the 'ok' value is handled

Worth noting - it was very hard to test this via the existing Bake tests. Because of this, we needed to export the function to be able to test it properly, but we know this is probably not ideal.

Our hope is you can help to review this and find a better way to test via the existing fixturing, or be able to confirm that making it public is acceptable

### DIFF
--- a/internal/commands/bake.go
+++ b/internal/commands/bake.go
@@ -437,7 +437,7 @@ func (b Bake) Execute(args []string) error {
 		// TODO remove when stemcell tarball is deprecated
 		stemcellManifest, err = b.stemcell.FromTarball(b.Options.StemcellTarball)
 	} else if b.Options.Kilnfile != "" {
-		if err := bakeArgumentsFromKilnfileConfiguration(&b.Options, templateVariables); err != nil {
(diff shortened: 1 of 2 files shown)


crhntr (approved):
This looks good. I like the test boundary. Exporting that function is the right move.

Most of the tests are passing locally for me. There is a broken acceptance tests. I don't think the failure is related to this change. I feel okay merging. I'll rebase #481 on main once this is merged and dig into fixing the assertion there. It might already be fixed.

crhntr:
@pabloarodas and I reviewed this and merged it as a pair.

PR #485, opened 2024-03-20
fix: Updated logic on when to bake Kilnfile arguments

ram-pivot:
- Bake arguments from kilnfile only when the stub releases flag is not set.
- This is causing TAS PRs checks to fail as currently for stub releases arguments are not needed.


crhntr (approved):
The docs look good to me. I just ran an auto formatter to fix the import paths.

pabloarodas (approved):

pabloarodas:
Installed version locally and it works now to generate a tile with `--stub-releases` and `--metadata-only`

PR #490, opened 2024-04-25
Use basic auth for all request to artifactory if credentials are defined

### LINKED ISSUE #474: Artifactory release source assumes the artifactory has anonymous access enabled
### Describe the bug

When using Artifactory release source, Kiln assumes the artifactory has anonymous access enabled for discovering and fetching release tarballs.

Current versions of artifactory oss have anonymous access disabled by default.

### Reproduction steps

1. Specify an artifactory release source pointing to an artifactory that has anonymous access disabled.
2.  Attempt any of following:
    * `kiln find-release-version ...`
    * `kiln bake ...`

### Expected behavior

The supplied credentials for the artifactory release source would be used for all requests to artifactory.

### Additional context

The artifactory release source in Kiln only uses authentication for the : [UploadRelease](https://github.com/pivotal-cf/kiln/blob/c526912b1cc10fe0127f36ab9b78ca48be0f6e74/internal/component/artifactory.go#L300) method

A current workaround is to add credentials to the `artifactory_host` url.  Ex:  
```
artifactory_host: https://username:password@build-artifactory.your-artifactory-url.com
```

dlinsley:
Resolves #474

verified local tests covers change.

tested following subcommands against an artifactory that required authentication:
- find-release-version
- update-stemcell
- update-release

### DIFF
--- a/internal/component/artifactory.go
+++ b/internal/component/artifactory.go
@@ -80,9 +80,9 @@ func (ars *ArtifactoryReleaseSource) DownloadRelease(releaseDir string, remoteRe
 	downloadURL += "/" + ars.Repo + "/" + remoteRelease.RemotePath
 
 	ars.logger.Printf(logLineDownload, remoteRelease.Name, ReleaseSourceTypeArtifactory, ars.ID)
-	resp, err := ars.Client.Get(downloadURL)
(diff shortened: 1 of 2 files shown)


dlinsley:
Will re-create this without fork

PR #491, opened 2024-04-25
use basic auth for all request to artifactory if credentials are defined

dlinsley:
Resolves https://github.com/pivotal-cf/kiln/issues/474

verified local tests covers change.

Tested following subcommands against an artifactory that required authentication:
- find-release-version
- update-stemcell
- update-release


crhntr (approved):
Looks good to me.

dlinsley:
Also confirmed this is functional for :
- `bake`  with & without `--final`
- `re-bake`

With an artifactory requiring authentication

pabloarodas (approved):

Issue #492, opened 2024-06-04
add `--without-download` to `update-stemcell` command if possible


dlinsley:
### Is your feature request related to a problem? Please describe.

With `kiln update-release` we have an option: `--without-download` which will just check the existence and checksum of releases in the remote release source

`kiln update-stemcell` does not have this option currently and requires downloading all of the respective releases for the new stemcell to determine the checksum of each compiled release against the new stemcell.

### Describe the solution you'd like

By having `kiln update-stemcell --without-download`  we can check the release source for existence and checksum of the compiled release without downloading them when compiled releases for a new stemcell are available.

dlinsley:
Alternative is to not add additional option, but update the release sources which support determining checksum of release files remotely to not download the file.  In this case: artifactory.  Not all release sources will have this capability.

dlinsley:
a PR implementing this functionality:  #493

Issue #297, opened 2021-11-11
Use `digest` instead of sha1 in Kilnfile


crhntr:
## Acceptance Criteria

A

The modified field NBF 
```diff
- digest                ::= algorithm ":" encoded
+ digest                ::= algorithm ":" encoded | encoded
  algorithm             ::= algorithm-component (algorithm-separator algorithm-component)*
  algorithm-component   ::= [a-z0-9]+
  algorithm-separator   ::= [+._-]
  encoded               ::= [a-zA-Z0-9=_-]+
```

dtimm:
This looks like a good change to me. The current requirement of Kiln downloading releases in many circumstances causes a lot of headaches in automation, and any reductions we can make in that are good. Deprecating and removing the old field in Kiln 1.0 would be a good, natural place to make the switch, as well.

PR #493, opened 2024-06-04
for update-stemcell, only download releases if necessary 

dlinsley:
implements #492 

* Only download releases if determining the SHA1 remotely was not possible.
* Updates artifactory `GetMatchedRelease` to include remote SHA1 of matched release from remote
* Updates bosh.io `GetMatchedRelease` to include remote SHA1 of matched release from remote
* Other release sources do not appear to support determining the remote SHA1.

### DIFF
--- a/internal/commands/update_stemcell.go
+++ b/internal/commands/update_stemcell.go
@@ -97,16 +97,19 @@ func (update UpdateStemcell) Execute(args []string) error {
 			update.Logger.Printf("No change for release %q\n", rel.Name)
 			continue
 		}
-
-		local, err := releaseSource.DownloadRelease(update.Options.ReleasesDir, remote)
-		if err != nil {
-			return fmt.Errorf("while downloading release %q, encountered error: %w", rel.Name, err)
-		}
-
 		lock := &kilnfileLock.Releases[i]
-		lock.SHA1 = local.Lock.SHA1
+
 		lock.RemotePath = remote.RemotePath
 		lock.RemoteSource = remote.RemoteSource
+		lock.SHA1 = remote.SHA1
+		if remote.SHA1 == "" || remote.SHA1 == "not-calculated" {
+			// release source needs to download.
+			local, err := releaseSource.DownloadRelease(update.Options.ReleasesDir, remote)
+			if err != nil {
+				return fmt.Errorf("while downloading release %q, encountered error: %w", rel.Name, err)
+			}
+			lock.SHA1 = local.Lock.SHA1
+		}
 	}
 
 	kilnfileLock.Stemcell.Version = trimmedInputVersion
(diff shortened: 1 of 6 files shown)


drich10 (changes requested):
Change looks good and I agree that defaulting this behavior rather than requiring a flag is correct.

Afaict there isn't a test that verifies the behavior change in update_stemcell.go itself. If I'm right, would you mind adding one?

internal/commands/update_stemcell.go
```diff
@@ -97,16 +97,19 @@ func (update UpdateStemcell) Execute(args []string) error {
...
-			return fmt.Errorf("while downloading release %q, encountered error: %w", rel.Name, err)
-		}
-
 		lock := &kilnfileLock.Releases[i]
-		lock.SHA1 = local.Lock.SHA1
+
 		lock.RemotePath = remote.RemotePath
 		lock.RemoteSource = remote.RemoteSource
+		lock.SHA1 = remote.SHA1
+		if remote.SHA1 == "" || remote.SHA1 == "not-calculated" {
```
crhntr, line 105: Where is "not-calculated" from? Maybe pull this out into a const and add a note on what emits this value.

(It was cool to see the PR review request. I miss y'all. Hope you're all well.)

crhntr (approved):
I added a comment. I don't recall anything that would break from this change.

This issue from 2021 is kinda related. We only calcuate the SHA1 because we used to have a huge s3 bucket that did not have object checksums enabled and adding it would have been super expensive from a network perspective. Funnily, if we had enabled that feature we would have saved thousands of release downloads.

https://github.com/pivotal-cf/kiln/issues/297

internal/commands/update_stemcell.go
dlinsley: The `s3` and `github` sources can set this as  [`not-calculated`](https://github.com/search?q=repo%3Apivotal-cf%2Fkiln%20not-calculated&type=code) particularly when no-download specified.

dlinsley:
> Change looks good and I agree that defaulting this behavior rather than requiring a flag is correct.
> 
> Afaict there isn't a test that verifies the behavior change in update_stemcell.go itself. If I'm right, would you mind adding one?

The updated tests add a third release to the tile under test that provides sufficient metadata to the update-stemcell method to not downloaded the release. Updated the returned metadata of the existing releases in the tile to explicitly return either empty string or `not-calculated` to trigger a calls to download the releases.

rizwanreza:
This broke our pipelines. It is not downloading the releases in our compilation manager. 

I am going to revert this PR and re-release kiln.

dlinsley:
Sorry about that, didn't realize the download was desired for this command, may put this behind a --no-download flag instead

PR #501, opened 2024-08-09
Kiln validate resource type allow list

jajita:
Golden Path requires tile authors to only use artifactory. This new flag allows us to upstream the requirement to Kiln. 

This doesn't introduce any breaking changes; new behavior is effectively feature flagged.

Story Id: TPCF-10575

### DIFF
--- a/pkg/cargo/validate.go
+++ b/pkg/cargo/validate.go
@@ -7,8 +7,40 @@ import (
 	"github.com/Masterminds/semver/v3"
 )
 
-func Validate(spec Kilnfile, lock KilnfileLock) []error {
+type ValidationOptions struct {
+	resourceTypeAllowList []string
+}
+
+func ValidateResourceTypeAllowList(allowList ...string) ValidationOptions {
+	return ValidationOptions{}.SetValidateResourceTypeAllowList(allowList)
+}
+
+func (o ValidationOptions) SetValidateResourceTypeAllowList(allowList []string) ValidationOptions {
+	o.resourceTypeAllowList = allowList
+	return o
+}
+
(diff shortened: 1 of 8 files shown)


pkg/cargo/validate.go
```diff
@@ -2,14 +2,51 @@ package cargo
...
+func (o ValidationOptions) SetValidateResourceTypeAllowList(allowList []string) ValidationOptions {
+	o.resourceTypeAllowList = allowList
+	return o
+}
+
+func mergeOptions(options []ValidationOptions) ValidationOptions {
+	var opt ValidationOptions
+	for _, o := range options {
+		if o.resourceTypeAllowList != nil {
+			opt.resourceTypeAllowList = o.resourceTypeAllowList
```
jajita, line 32: Should we append to the list instead of replacing?

crhntr: The allow list already allows multiple options. I suspect having the call site reset the state of the list to just the parameter might be easier to track what is actually allowed.

pvaramballypivot:
I just want to confirm that you need to pass all release source types for kiln to validate the file if you pass the new flag.

`../kiln/kiln validate -kf tas/Kilnfile -vr github_access_token=ed -vr artifactory_username=foo -vr artifactory_password=bar --allow-release-source-type=github --allow-release-source-type=[bosh.io](http://bosh.io/) --allow-release-source-type=artifactory`

Meaning, if I just pass one of them, lets say github, should it throw an error that artifactory and bosh.io are not allowed? That's what it is doing now.

Please confirm and il merge the PR

PR #503, opened 2024-08-12
Update Kiln  to Go 1.22.6

bjanice75:
Updating the Go version in Kiln to 1.22.6


crhntr (approved):
I'm glad we are getting ahead of 1.21 going end of support sometime in the next few weeks when 1.23 is released.

You don't need to bump the versions in the test fixtures. The Go toolchain will run in a sort of compatibility mode when the version in the Go module is older than the toolchain in use. If you want to make the change, just copy paste the checksum values from the error into the test assertions.

internal/acceptance/workflows/testdata/tiles/v2/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/crhntr/hello-tile
 
-go 1.20
```
crhntr, line 3: As far as I know, you don't need to bump this. When run with a newer version, Go should just run in a sort of compatibility mode.

bjanice75: Thanks Chris.  Will you be merging this?

PR #506, opened 2024-08-22
Bumps ruby version to 3.2.5

ryanwittrup:
Experienced issues when trying to run kiln test - investigation showed a failure during the step to build the dockerfile based on a CGO error:

```
usr/lib/gcc/x86_64-linux-gnu/10/cc1: error while loading shared libraries: usr/lib/x86_64-linux-gnu/libisl.so.23: file too short
```

One approach was to disable CGO, via CGO_ENABLED=0

This ruby bump felt like a better approach, but CGO could be considered as part of the PR review if there are concerns related to compatibility between the new ruby patch and the pinned OpsMan version

[TPCF-24856](https://jira.eng.vmware.com/browse/TPCF-24856)

Authored-by: Ryan Wittrup <rwittrup@vmware.com>

### DIFF
--- a/internal/test/Dockerfile
+++ b/internal/test/Dockerfile
@@ -3,11 +3,11 @@
 FROM golang as go-image
 FROM pivotalcfreleng/kiln:v0.77.0 as kiln
 
-FROM ruby:3.2.0 as builder
+FROM ruby:3.2.5 as builder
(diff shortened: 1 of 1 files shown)


ryanwittrup:
Closing for now - may not be necessary since it seems like my local failures were related to a corrupted docker cache

Got the dockerfile build working again with existing code by pruning everything
```
docker system prune --all
```

PR #510, opened 2024-10-03
Converge on the new version of Go Github

crhntr:
bump go-github from v50 and v40 to v65


crhntr:
Never mind :) we will do this in our branch so we can do the manual testing once.

PR #511, opened 2024-11-11
Reduce memory usage when calculating release shasum

geofffranks:
Reads in boshreleases a chunk at a time + does a streaming sha1sum while discarding the memory, rather than reading the entirety of the boshrelease into memory and then discarding.


davewalter (approved):
Derek ran the tests locally and confirmed that they passed.

PR #515, opened 2025-01-06
do the sha1 hash during download instead of after to reduce local I/O…

dlinsley:
… for artifactory and bosh.io release sources.

Already implemented for github release source.

Incompatible with the s3 release source which uses a concurrent download client.

### DIFF
--- a/internal/component/artifactory_release_source.go
+++ b/internal/component/artifactory_release_source.go
@@ -100,23 +100,15 @@ func (ars *ArtifactoryReleaseSource) DownloadRelease(releaseDir string, remoteRe
 	if err != nil {
 		return Local{}, err
 	}
-	defer func() { _ = out.Close() }()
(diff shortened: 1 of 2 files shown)


internal/component/artifactory_release_source.go
```diff
@@ -100,25 +100,17 @@ func (ars *ArtifactoryReleaseSource) DownloadRelease(releaseDir string, remoteRe
 	if err != nil {
 		return Local{}, err
 	}
-	defer func() { _ = out.Close() }()
+	defer closeAndIgnoreError(out)
```
crhntr, line 103: Happy to see my `closeAndIgnoreError` pattern sticking around :)

crhntr (approved):
I suspect y'all will save a few seconds per tile in CI with this change. Nice!

PR #518, opened 2025-01-10
Improve debug output

davewalter:
- **Add the remote version to the download log message**
- **Remove the file info log line from the Artifactory release source**


internal/component/artifactory_release_source.go
```diff
@@ -126,7 +126,6 @@ func (ars *ArtifactoryReleaseSource) DownloadRelease(releaseDir string, remoteRe
 
 func (ars *ArtifactoryReleaseSource) getFileSHA1(release cargo.BOSHReleaseTarballLock) (string, error) {
 	fullURL := ars.ArtifactoryHost + "/api/storage/" + ars.Repo + "/" + release.RemotePath
-	ars.logger.Printf("Getting %s file info from artifactory", release.Name)
```
drich10, line 129: @davewalter was this just being noisy?

drich10 (approved):
Looks like a useful additive improvement.

The other option I would have considered is listing out the remote url(s) based on the source as well or instead?

davewalter:
> The other option I would have considered is listing out the remote url(s) based on the source as well or instead?

I could see that being helpful for debugging, but maybe too noisy for verification that it is doing what you expect?

PR #519, opened 2025-01-11
Add a without-download flag to update-stemcell to skip release downloads

davewalter:
This PR re-applies the changes from #493 with the addition of a without-download
flag to control the behaviour.

### DIFF
--- a/internal/commands/update_release.go
+++ b/internal/commands/update_release.go
@@ -16,11 +16,11 @@ type UpdateRelease struct {
 	Options struct {
 		flags.Standard
 
-		Name                         string `short:"n" long:"name" required:"true" description:"name of release to update"`
-		Version                      string `short:"v" long:"version" required:"true" description:"desired version of release"`
-		ReleasesDir                  string `short:"rd" long:"releases-directory" default:"releases" description:"path to a directory to download releases into"`
-		AllowOnlyPublishableReleases bool   `long:"allow-only-publishable-releases" description:"include releases that would not be shipped with the tile (development builds)"`
-		WithoutDownload              bool   `long:"without-download" description:"updates releases without downloading them"`
+		Name                         string `short:"n"  long:"name"                            required:"true"    description:"name of release to update"`
+		Version                      string `short:"v"  long:"version"                         required:"true"    description:"desired version of release"`
+		ReleasesDir                  string `short:"rd" long:"releases-directory"              default:"releases" description:"path to a directory to download releases into"`
+		AllowOnlyPublishableReleases bool   `           long:"allow-only-publishable-releases" default:"false"    description:"include releases that would not be shipped with the tile (development builds)"`
+		WithoutDownload              bool   `short:"wd" long:"without-download"                default:"false"    description:"updates releases without downloading them"`
 	}
 	multiReleaseSourceProvider MultiReleaseSourceProvider
 	filesystem                 billy.Filesystem
--- a/internal/commands/update_stemcell.go
+++ b/internal/commands/update_stemcell.go
@@ -19,9 +19,10 @@ type UpdateStemcell struct {
 	Options struct {
 		flags.Standard
 
-		Version        string `short:"v"   long:"version"               required:"true"    description:"desired version of stemcell"`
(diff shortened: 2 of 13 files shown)


internal/commands/update_stemcell.go
```diff
@@ -12,16 +12,16 @@ import (
...
 	Options struct {
 		flags.Standard
 
-		Version        string `short:"v"   long:"version"               required:"true"    description:"desired version of stemcell"`
-		ReleasesDir    string `short:"rd"  long:"releases-directory"    default:"releases" description:"path to a directory to download releases into"`
-		UpdateReleases bool   `short:"ur"  long:"update-releases"                          description:"finds latest matching releases for new stemcell version"`
+		Version         string `short:"v"   long:"version"               required:"true"    description:"desired version of stemcell"`
+		ReleasesDir     string `short:"rd"  long:"releases-directory"    default:"releases" description:"path to a directory to download releases into"`
+		UpdateReleases  bool   `short:"ur"  long:"update-releases"                          description:"finds latest matching releases for new stemcell version"`
+		WithoutDownload bool   `            long:"without-download"                         description:"updates stemcell releases without downloading releases"`
```
drich10, line 24: maybe add a short since all the other options have one? `wd` seems like it would fit the pattern to me

drich10 (approved):
Looks good to me, just one suggestion to add a short flag for the new option

internal/commands/update_stemcell.go
```diff
@@ -69,44 +69,48 @@ func (update UpdateStemcell) Execute(args []string) error {
...
-		var remote cargo.BOSHReleaseTarballLock
-		if update.Options.UpdateReleases {
-			remote, err = releaseSource.FindReleaseVersion(spec, true)
-		} else {
+		if !update.Options.UpdateReleases {
 			spec.Version = rel.Version
-			remote, err = releaseSource.GetMatchedRelease(spec)
 		}
 
+		remote, err := releaseSource.FindReleaseVersion(spec, true)
```
dlinsley, line 86: `FindReleaseVersion` could return a newer version than desired since I don't believe all implementations use `spec.Version` if it is specified

In particular, I noticed the artifactory release source only  uses `spec.StemcellVersion` and the version constraint from the Kilnfile `spec.VersionConstraint` with FindReleaseVersion.

Before the --update-releases` option is added, this code path only used releaseSource.GetMatchedRelease

davewalter: Thanks @dlinsley. I was unsure how to proceed here. I assume you are thinking that this would work better?

```golang
if !update.Options.UpdateReleases {
  spec.Version = rel.Version
  remote, err = releaseSource.GetMatchedRelease(spec)
} else {
  remote, err = releaseSource.FindReleaseVersion(spec, true)
}
```

dlinsley: Yes, I thought that would be more predictable.

internal/commands/update_release.go
```diff
@@ -16,11 +16,11 @@ type UpdateRelease struct {
...
-		AllowOnlyPublishableReleases bool   `long:"allow-only-publishable-releases" description:"include releases that would not be shipped with the tile (development builds)"`
-		WithoutDownload              bool   `long:"without-download" description:"updates releases without downloading them"`
+		Name                         string `short:"n"  long:"name"                            required:"true"    description:"name of release to update"`
+		Version                      string `short:"v"  long:"version"                         required:"true"    description:"desired version of release"`
+		ReleasesDir                  string `short:"rd" long:"releases-directory"              default:"releases" description:"path to a directory to download releases into"`
+		AllowOnlyPublishableReleases bool   `           long:"allow-only-publishable-releases" defailt:"false"    description:"include releases that would not be shipped with the tile (development builds)"`
+		WithoutDownload              bool   `short:"wd" long:"without-download"                defailt:"false"    description:"updates releases without downloading them"`
```
dlinsley, line 23: nit: should `defilt` be ``default`?

davewalter: Yes, it should. Fixing...

davewalter:
@dlinsley I've added some acceptance tests to cover the `--update-releases`  and`--without-download` flags for `update-release` and `update-stemcell`. Could you please take another look?

dlinsley (approved):

Issue #516, opened 2025-01-07
path template may not match all of a artifact name when similar artifact is present


dlinsley:
### Describe the bug

Using an artifactory source

Files that do not fully match the path template can be recommended if they sufficiently match the `{{.Name}}` and `{{.Version}}` fields.

I do not know if this is artifactory release source specific yet,

### Reproduction steps

Using Kiln 0.99.0

1. have a Kiln file reference an artifactory release source where the path template expects compiled bosh releases
ex:  bosh-releases/compiled/{{.Name}}-{{.Version}}-{{.StemcellOS}}-{{.StemcellVersion}}.tgz
2.  have release source (bosh-releases/compiled) have following files where StemcellOS: smoothie, StemcellVersion: 9.9:
  - `mango-2.3.4-notices.zip`
  - `mango-2.3.4-smoothie-9.9.tgz`
3. `kiln find-release-version -r mango` would return `2.3.4`
4. `kiln update-release -n mango --version 2.3.4`  updates the Kilnfile.lock

### Expected behavior

Kilnfile.lock should reference `mango-2.3.4-smoothie-9.9.tgz` but instead references: `mango-2.3.4-notices.zip`

Expected the all of the path template to match.  But it looks like it just matched {{.Name}} and {{.Version}}, and ignored the stemcell parameters and suffix `.tgz`

### Additional context

Updated tests that exhibit the behavior available on branch :  [artifactory-path-template-tests](https://github.com/pivotal-cf/kiln/tree/artifactory-path-template-tests)

PR #533, opened 2025-03-10
Find Release Version - Fix Inexact Artifactory Remote Matching

notrepo05:


notrepo05:
As part of validating this change I setup an Artifactory folder with two entries in order to reproduce the bug with a variant of `hello-tile`:

```
hello-release-1.0.0-ubuntu-jammy-1.737.tgz
hello-release-1.0.5-notices.zip
```

Running `find-release-version` on `hello-release` prior to the update:

```
kiln find-release-version -r hello-release
{"version":"1.0.5","remote_path":"compiled-releases/hello-release/hello-release-1.0.5-notices.zip","source":"artifactory","sha":"0f237c64502c6acfe6e6dcf7be43da569833903e"}
```

and after we see the correctly discovered hello-release file:

```
go -C ~/workspace/kiln run main.go find-release-version -r hello-release
{"version":"1.0.5","remote_path":"compiled-releases/hello-release/hello-release-1.0.5-ubuntu-jammy-1.737.tgz","source":"artifactory","sha":"0f237c64502c6acfe6e6dcf7be43da569833903e"}
```

dlinsley:
fixes #516

internal/component/artifactory_release_source.go
```diff
@@ -247,15 +247,38 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
...
+		remotePathFilename := filepath.Base(remotePath)
+
+		// we want to ensure that the path template and the artifactory file being considered
+		// having matching formats. we strip the versions out because those aren't material
+		// to this check
+		if removeVersions.Replace(remotePathFilename) != removeVersions.Replace(artifactoryFilename) {
+			continue
+		}
+
+		stemcellVersion := versionFormatter.Replace(versions[len(versions)-1])
```
dlinsley, line 272: This appears to make an assumption that the stemcell version is the last version in the pattern. Is that accurate?

internal/component/artifactory_release_source.go
```diff
@@ -247,15 +247,38 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
...
+			replacements = append(replacements, ver)
+			replacements = append(replacements, "")
+		}
+		removeVersions := strings.NewReplacer(replacements...)
+		remotePathFilename := filepath.Base(remotePath)
+
+		// we want to ensure that the path template and the artifactory file being considered
+		// having matching formats. we strip the versions out because those aren't material
+		// to this check
+		if removeVersions.Replace(remotePathFilename) != removeVersions.Replace(artifactoryFilename) {
```
notrepo05, line 268: I may need to add a test case for when we're moving from a compiled released to an uncompiled release

internal/component/artifactory_release_source.go
notrepo05: I believe it is, but it could be made clearer by only using it where it's valid. 

```
if len(versions) > 1 {
	stemcellVersion := versionFormatter.Replace(versions[len(versions)-1])
	if stemcellVersion != spec.StemcellVersion {
		continue
	}
}
```

reusing the existing convention of the last slice entry

internal/component/artifactory_release_source_test.go
```diff
@@ -88,6 +88,16 @@ var _ = Describe("interacting with BOSH releases on Artifactory", func() {
 				defer closeAndIgnoreError(f)
 				_, _ = io.Copy(res, f)
 			}), requireAuth))
+			artifactoryRouter.Handler(http.MethodGet, "/api/storage/basket/bosh-releases/smoothie/9.9/mango/mango-2.3.4-notices.zip", applyMiddleware(http.HandlerFunc(func(res http.ResponseWriter, _ *http.Request) {
+				res.WriteHeader(http.StatusOK)
+				// language=json
+				_, _ = io.WriteString(res, `{"checksums": {"sha1":  "some-sha-zip"}}`)
+			}), requireAuth))
+			artifactoryRouter.Handler(http.MethodGet, "/api/storage/basket/bosh-releases/smoothie/9.9/mango/mango-2.3.4-smoothie-9.9-notices.zip", applyMiddleware(http.HandlerFunc(func(res http.ResponseWriter, _ *http.Request) {
```
davewalter, line 96: Should this handler be responding to requests to get `/mango-2.3.4-smoothie-9.9-notices.tgz` to match the filename being returned on line 80?

davewalter (commented):
Makes sense overall. I just had one question about the changes to the Artifactory release source test setup.

internal/component/artifactory_release_source_test.go
notrepo05: Thanks @davewalter, that's correct and I pushed a commit fixing this.

dlinsley (approved):
looks great, thank you!

notrepo05:
@davewalter Should `find-release-version` support discovering pre-release versions? An example would be `some-release-1.2.3-build.1-stemcell-1.2.3.tgz`? Prior to this change they're discoverable but the Artifactory Release Source truncated the pre-release:

```
kiln find-release-version -r some-release \
   --vr artifactory_host=<redacted> \
   --vr artifactory_repo=<redacted> \
   --vr artifactory_username=<redacted> \
   --vr artifactory_password=<redacted>
{"version":"1.2.3","remote_path":"compiled-releases/some-release/some-release-1.2.3-build.1-ubuntu-jammy-1.2.3.tgz","source":"artifactory","sha":"abc"}
```

It seems like a bug to truncate the pre-release info in the `version` field?

rizwanreza:
Ran the acceptance test suite and confirmed this is working as expected.

PR #538, opened 2025-03-13
fix: variable file default discovery for nested tiles

notrepo05:


dlinsley (approved):
Thanks for the details in the PR description!

PR #543, opened 2025-03-31
Kiln test fixes

rkoster:
Reuse image if it already exists
Install ops-manifest from new canonical home
Bump kiln cli


rizwanreza:
My read is that it might not be relevant anymore. Happy to take another look if needed.

PR #547, opened 2025-04-15
Template into complex strings without quoting

### LINKED ISSUE #541: Version is templated with unexpected quoting
### Describe the bug

The `$( version )` template helper provides access to the version. When using it for a YAML value on its own, it renders the version as expected without quotes. However, when interpolating it into a more complex string value, then the result contains unexpected quotes around the version.

### Reproduction steps

1. Template metadata with a complex value construction 
    ```yaml
    something: hello-$( version )
    ```
2. Bake the tile by providing a version, say `--version 1.2.3`
3. Assert that the result has quotes inside
    ```yaml
    something: hello-"1.2.3"
    ```

### Expected behavior

1. Template metadata with a complex value construction 
    ```yaml
    something: hello-$( version )
    ```
2. Bake the tile by providing a version, say `--version 1.2.3`
3. Assert that the result has no quotes inside
    ```yaml
    something: hello-1.2.3
    ```

### Additional context

_No response_

mamachanko:
Fixes #541

Signed-off-by: Max Brauer <mbrauer@vmware.com>

### DIFF
--- a/internal/builder/interpolator.go
+++ b/internal/builder/interpolator.go
@@ -274,6 +274,10 @@ func (i Interpolator) interpolateValueIntoYAML(input InterpolateInput, name stri
 		return "", err // should never happen
 	}
 
+	if stringScalar, ok := val.(string); ok {
+		return stringScalar, nil
+	}
+
 	interpolatedYAML, err := i.interpolate(input, name, initialYAML)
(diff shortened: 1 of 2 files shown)


rizwanreza:
Ran acceptance test suite with this change.

PR #552, opened 2025-04-30
Revert "Find Release Version - Fix Inexact Artifactory Remote Matching"

rizwanreza:
Reverts pivotal-cf/kiln#533


rizwanreza:
cc @notrepo05

PR #553, opened 2025-04-30
Consume ops-manifest as a pre-built gem

davewalter:
Skip the integration test for the internal/test package as it requires
access to Artifactory.


rizwanreza:
Tests passed. Merging.

PR #556, opened 2025-05-01
Feat/artifactory source full semver support

notrepo05:
## Issue

`kiln` is extracting versions matching the following the pattern `([-v])\d+(.\d+)*` and assuming the first is the bosh release version while if a second version is present it represents the stemcell version. While this works for a wide variety of cases there are some edge cases where it's brittle. For example `some-release-1.2.3-notices.zip` is matched erroneously. Also, in the case a pre-release segment is wanted then the version extracted won't contain it, meaning the ordering of pre-releases isn't well defined. This can be seen by how the version is extracted:

```golang
		versions := semverPattern.FindAllString(filepath.Base(releases.URI), -1)
		version := versions[0]
```

and then sorted

```golang
newVersion, _ := semver.NewVersion(version)
...
foundVersion, _ := semver.NewVersion(foundRelease.Version)
if newVersion.GreaterThan(foundVersion) {
```

## Proposal

I'd like to propose that the artifactory source in particular is brought fully in line with the semver semantics used by mastermind's v3. So that `*` does not match pre-releases allowing tiles to avoid accidental ingestion of them, while still permitting matches when desired using the convention `>0.0.0-0`. See https://github.com/Masterminds/semver?tab=readme-ov-file#working-with-prerelease-versions

### DIFF
--- a/internal/component/artifactory_release_source.go
+++ b/internal/component/artifactory_release_source.go
@@ -232,8 +232,23 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
 		return cargo.BOSHReleaseTarballLock{}, fmt.Errorf("json from %s is malformed: %s", response.Request.URL.Host, err)
 	}
 
-	semverPattern, err := regexp.Compile(`([-v])\d+(.\d+)*`)
+	semverRegex := `v?(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?` +
+		`(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?` +
+		`(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
+
+	regexSpec := spec
+	regexSpec.Version = fmt.Sprintf(`(?P<bosh_version>(%s))`, semverRegex)
+	regexSpec.StemcellVersion = fmt.Sprintf(`(?P<bosh_stemcell_version>(%s))`, semverRegex)
+	semverFilepathRegex, err := ars.RemotePath(regexSpec)
 	if err != nil {
+		return cargo.BOSHReleaseTarballLock{}, err
+	}
+	semverFilepathRegex = filepath.Base(semverFilepathRegex)
+	fmt.Println(semverFilepathRegex)
+
+	re, err := regexp.Compile(semverFilepathRegex)
+	if err != nil {
+		fmt.Println("failed TO COMPILE SEMVER PATTERN:", re)
 		return cargo.BOSHReleaseTarballLock{}, err
 	}
 
@@ -247,18 +262,38 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
 		if releases.Folder {
 			continue
 		}
-		versions := semverPattern.FindAllString(filepath.Base(releases.URI), -1)
-		version := versions[0]
-		stemcellVersion := versions[len(versions)-1]
-		version = strings.Replace(version, "-", "", -1)
-		version = strings.Replace(version, "v", "", -1)
-		stemcellVersion = strings.Replace(stemcellVersion, "-", "", -1)
-		if len(versions) > 1 && stemcellVersion != spec.StemcellVersion {
+		// mango-2.3.4-build.1-smoothie-9.9.tgz
+		// mango-?P<version>(v?(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)-smoothie-9.9.tgz
+		m := re.FindStringSubmatch(filepath.Base(releases.URI))
+		if m == nil {
 			continue
 		}
+		names := re.SubexpNames()
+		data := map[string]string{}
+		for i, n := range names {
+			if i == 0 || n == "" {
+				continue
+			}
+			data[n] = m[i]
+		}
+
+		version := data["bosh_version"]
+		stemcellVersion := data["bosh_stemcell_version"]
+		if stemcellVersion != spec.StemcellVersion {
+			continue
+		}
+
+		//stemcellVersion := versions[len(versions)-1]
+		//version = strings.Replace(version, "-", "", -1)
+		//version = strings.Replace(version, "v", "", -1)
+		//stemcellVersion = strings.Replace(stemcellVersion, "-", "", -1)
+		//if len(versions) > 1 && stemcellVersion != spec.StemcellVersion {
+		//	continue
+		//}
 		if version != "" {
 			newVersion, _ := semver.NewVersion(version)
-			if !constraint.Check(newVersion) {
+			check := constraint.Check(newVersion)
+			if !check {
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


notrepo05:
This is intended to supersede https://github.com/pivotal-cf/kiln/pull/533 which was reverted to maintain compatibility with pre release versions

notrepo05:
Notes:

* The leading `v` isn't currently removed from remotely discovered versions, but this doesn't impact finding new versions. I need to work out what to do in this case
* When a stemcell os is configured but the version isn't it's treated as an error.
* This will require tiles using pre-releases to adopt the new pre-release constraint

internal/component/artifactory_release_source.go
```diff
@@ -247,18 +267,33 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
 		if releases.Folder {
 			continue
 		}
-		versions := semverPattern.FindAllString(filepath.Base(releases.URI), -1)
-		version := versions[0]
-		stemcellVersion := versions[len(versions)-1]
-		version = strings.Replace(version, "-", "", -1)
-		version = strings.Replace(version, "v", "", -1)
-		stemcellVersion = strings.Replace(stemcellVersion, "-", "", -1)
-		if len(versions) > 1 && stemcellVersion != spec.StemcellVersion {
```
notrepo05, line 256: Replace parsing with a regex constructed from the template

notrepo05:
I ran git bisect and discovered that 3b4fd72f6c2a5d2a8576a60bef0708a4c3d6ac0f might be breaking the bake acceptance tests. I've reverted it for now in order to have a clean test foundation

notrepo05:
Tests pass otherwise, so reapplying with that known. That should be followed up on in a different PR :)

internal/component/artifactory_release_source.go
```diff
@@ -232,7 +232,27 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
 		return cargo.BOSHReleaseTarballLock{}, fmt.Errorf("json from %s is malformed: %s", response.Request.URL.Host, err)
 	}
 
-	semverPattern, err := regexp.Compile(`([-v])\d+(.\d+)*`)
+	// https://github.com/Masterminds/semver/blob/master/version.go
```
mamachanko, line 235: **[suggestion, non-blocking]:** use a permalink.

internal/component/artifactory_release_source.go
```diff
@@ -247,18 +267,33 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
...
-		versions := semverPattern.FindAllString(filepath.Base(releases.URI), -1)
-		version := versions[0]
-		stemcellVersion := versions[len(versions)-1]
-		version = strings.Replace(version, "-", "", -1)
-		version = strings.Replace(version, "v", "", -1)
-		stemcellVersion = strings.Replace(stemcellVersion, "-", "", -1)
-		if len(versions) > 1 && stemcellVersion != spec.StemcellVersion {
+		// mango-2.3.4-build.1-smoothie-9.9.tgz
+		// mango-(?P<version>{regex pattern})-smoothie-(?P<stemcell_version>{regex_pattern}).tgz
+		// {Name}-(?P<version>{regex pattern})-{StemcellOS}-(?P<stemcell_version>{regex_pattern}).tgz
```
mamachanko, line 272: **[suggestion, non-blocking]:** express a bit more intent with this comment

internal/component/artifactory_release_source_test.go
```diff
@@ -66,91 +67,295 @@ var _ = Describe("interacting with BOSH releases on Artifactory", func() {
...
+					}
+					type ApiStorageListing struct {
+						Children []ApiStorageChildren `json:"children"`
+					}
+
+					apiStorageListing := ApiStorageListing{}
+					for _, filename := range []string{
+						"mango-2.3.4-build.1-smoothie-9.9.tgz",
+						"mango-2.3.4-smoothie-9.9.tgz",
+						"mango-2.3.4-build.2-smoothie-9.9.tgz",
```
mamachanko, line 87: ~**[question, non-blocking]:** The test says "when there are pre-releases and full releases" but here we only add pre-releases. What I am I missing?~ Scratch that. I get it.

internal/component/artifactory_release_source_test.go
```diff
@@ -66,91 +67,295 @@ var _ = Describe("interacting with BOSH releases on Artifactory", func() {
...
+				When("we allow pre-releases", func() {
+					It("finds the latest version", func() { // testing FindReleaseVersion
+						//s := cargo.BOSHReleaseTarballSpecification{
+						//	Name:             "test",
+						//	Version:          "~2.0",
+						//	GitHubRepository: "git@github.com:test-org/test.git",
+						//}
+						resultLock, resultErr := source.FindReleaseVersion(cargo.BOSHReleaseTarballSpecification{
+							Name:            "mango",
+							Version:         ">=0.0.0-build.0",
```
mamachanko, line 131: **[suggestion, non-blocking]:** use `>0.0.0-0` as the smallest possible prerelease constraint to not overfit against `build`.

internal/component/artifactory_release_source_test.go
mamachanko: **[question]:** shouldn't we also add a prerelease that's newer than `2.3.4`, say `3.0.0-build.1`, and test that it gets retrieved?

internal/component/artifactory_release_source.go
```diff
@@ -232,7 +232,27 @@ func (ars *ArtifactoryReleaseSource) FindReleaseVersion(spec cargo.BOSHReleaseTa
 		return cargo.BOSHReleaseTarballLock{}, fmt.Errorf("json from %s is malformed: %s", response.Request.URL.Host, err)
 	}
 
-	semverPattern, err := regexp.Compile(`([-v])\d+(.\d+)*`)
+	// https://github.com/Masterminds/semver/blob/master/version.go
+	semverRegex := `v?(0|[1-9]\d*)(?:\.(0|[1-9]\d*))?(?:\.(0|[1-9]\d*))?` +
+		`(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?` +
+		`(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?`
```
mamachanko, line 238: **[question, non-blocking]:** I don't have the full context, but what limits us from using Mastermind's semver instead of regular expressions?

mamachanko (approved):

internal/component/artifactory_release_source_test.go
notrepo05: Good catch, I updated the test

internal/component/artifactory_release_source_test.go
notrepo05: Thanks, I did so

internal/component/artifactory_release_source.go
notrepo05: This comment pointed towards the the need to refactor, so I extracted the regex building into a method and removed the comment

internal/component/artifactory_release_source.go
notrepo05: I don't think they export the their regex. The idea is to extract exactly what they think qualifies as a semantic version and then pass that along for parsing to their library. I'm open to suggestions!

internal/component/artifactory_release_source.go
```diff
@@ -44,6 +44,7 @@ type ArtifactoryFolderInfo struct {
 		URI    string `json:"uri"`
 		Folder bool   `json:"folder"`
 	} `json:"children"`
+	Path string `json:"path"`
```
notrepo05, line 47: I introduced path so we ensure that the entire path is being matched on. For example
```json
{
  "repo" : "a-repo-dev-local",
  "path" : "/compiled-releases/release",
  "created" : "2024-04-23T22:44:56.231Z",
  "createdBy" : "some-user",
  "lastModified" : "2024-04-23T22:44:56.231Z",
  "modifiedBy" : "some-user",
  "lastUpdated" : "2024-04-23T22:44:56.231Z",
  "children" : [ {
    "uri" : "/release-0.102.0-ubuntu-jammy-1.406.tgz",
    "folder" : false
  }, {
    "uri" : "/release-0.103.0-ubuntu-jammy-1.406.tgz",
    "folder" : false
  },
...
}
```

should not only look at the children's uris but also the path `/compiled-releases/release`, which must match as well. Including if it includes stemcell or version info

mamachanko:
> I ran git bisect and discovered that [3b4fd72](https://github.com/pivotal-cf/kiln/commit/3b4fd72f6c2a5d2a8576a60bef0708a4c3d6ac0f) might be breaking the bake acceptance tests. I've reverted it for now in order to have a clean test foundation

https://github.com/pivotal-cf/kiln/issues/559 captures the regression and a fix

PR #639, opened 2026-04-15
feat(carvel): generate unique bosh release names

notrepo05:
Appends a SHA-256 content hash of .boshrelease inputs to the release version (e.g., 10.4.0+a1b2c3d4e5f6) so that different builds produce distinct versions preventing Artifactory collisions and BOSH Director upload errors

feat(carvel bake): uses the release version for the destination tarball path

### DIFF
--- a/internal/carvel/baker.go
+++ b/internal/carvel/baker.go
@@ -1,14 +1,18 @@
 package carvel
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


internal/carvel/baker.go
```diff
@@ -666,29 +681,91 @@ func (b *baker) createBoshRelease() error {
...
 	b.log("output: " + string(out))
 	if err != nil {
 		return err
 	}
 
+	b.progress(fmt.Sprintf("  BOSH release version: %s", releaseVersion))
 	return nil
 }
 
+func hashBoshReleaseInputs(boshReleaseDir string) (string, error) {
```
notrepo05, line 718: we can't use the sha of the bosh release without building it twice, but still prefer to have something deterministic and reproducible. the alternative could simply be a generated uuid.

rizwanreza:
Looking good on first read.. happy to review it again when it's final.

rizwanreza (approved):
LGTM!

PR #643, opened 2026-04-30
Route kiln test through docker-virtual registry and refactor test-script generation

rizwanreza:
## Summary

- Route `kiln test` image pulls through the internal Broadcom Artifactory docker-virtual registry, eliminating the need for a separate `docker login` before running tile tests.
- Rebuild the test-script generation layer (`testPlan` / `suiteStep`) so each suite runs in its own subshell with an independent exit code, colored pass/fail summary, and optional per-suite timestamps.
- Add `--verbose` / `--silent` / `--stability` flags; quiet npm by default; set `GOMAXPROCS` inside the container.
- Refactor for maintainability: decompose `runTest`, fix naming inconsistencies, and clean up small code smells.

### Dockerfile (`internal/test/Dockerfile`)

- All `FROM` lines now use `tas-rel-eng-docker-virtual.usw1.packages.broadcom.com` (the same registry as `AuthConfigs`) so `kiln test` can pull base images without a prior `docker login`.
- Stable tool layers (Go runtime, `jq`, `nodejs`, `npm`) are ordered before the `ARG ARTIFACTORY_USERNAME` declaration so they survive credential rotation.
- `ginkgo@v1.16.5` is pinned; `GOPROXY` credentials are scoped to a single `RUN` step (not exported as `ENV`) to keep that layer stable.
- `ops-manifest` gem install credentials are exported to `ENV` for runtime use.

### `kiln test` behaviour

- New flags: `--stability` (run only stability tests), `--silent` (suppress kiln info lines), `--verbose` (print container ID and per-suite timestamps).
- `--verbose` defaults to `false`; npm runs `--silent` / `ci --silent` by default and uses the louder form only when `--verbose` is passed.
- `GOMAXPROCS` is set to the host's CPU count inside the container.
- Credentials may be supplied via `-e ARTIFACTORY_USERNAME=... -e ARTIFACTORY_PASSWORD=...` or exported in the shell; `kiln test` exits with an error before contacting Docker if either is missing.

### `internal/test/container.go` refactors

- **`RunMetadata` → `RunStability`**: field name now matches the `--stability` flag and "Stability Tests" suite label.
- **Decompose `runTest`**: extracted `buildTestImage` (image build + registry auth) and `startAndWaitContainer` (container lifecycle + log drain + signal handling) so `runTest` is a 30-line orchestrator.
- **Single credential resolution**: `requiredArtifactoryCredentialsFromMap` reuses the already-decoded `environmentVars` map, eliminating a redundant `decodeEnvironment` call.
- **`verbose` as a render parameter**: removed `verbose` from the `testPlan` struct; `testPlan` is now pure data and `script(verbose bool)` makes the rendering concern explicit.
- **`GinkgoFlags` typo fixed**: `TileTest.Options.GingkoFlags` → `GinkgoFlags`.
- **`errors.New`**: replaced `fmt.Errorf("%s", detail)` in `checkImageBuildResponse`.
- **`filepath` consistency**: removed the `"path"` import; `path.Dir/Base` replaced with `filepath.Dir/Base` for host filesystem operations.

### README

Updated `kiln test` docs to reflect the new flags (`--stability`, `--ginkgo-flags`, `--manifest`, `--migrations`), the credential flow through `docker-virtual`, and the `DockerVirtualRegistryHost` alignment requirement.

#### `--verbose` flag

```
SUCCESS! -- 520 Passed | 0 Failed | 0 Pending | 4 Skipped


Ginkgo ran 1 suite in 8m46.345003494s
Test Suite Passed
[21:13:42] Completed: Manifest Tests

[21:03:23] ✓ Migration Tests Passed
[21:04:55] ✓ Stability Tests Passed
[21:13:42] ✓ Manifest Tests Passed
```

#### without `--verbose` flag

```
SUCCESS! -- 520 Passed | 0 Failed | 0 Pending | 4 Skipped


Ginkgo ran 1 suite in 8m10.665858116s
Test Suite Passed

✓ Migration Tests Passed
✓ Stability Tests Passed
✓ Manifest Tests Passed
```


internal/test/container.go
```diff
@@ -119,24 +135,49 @@ func runTest(ctx context.Context, logger *log.Logger, w io.Writer, dockerDaemon
...
+		SuppressOutput: true,
 	})
 
 	if err != nil {
 		return fmt.Errorf("failed to build image: %w", err)
 	}
 
-	logger.Println("reading image build response")
-	if err := checkImageBuildResponse(resp.Body); err != nil {
+	if err := checkImageBuildResponse(resp.Body, nil); err != nil {
```
notrepo05: nit: Can we refactor `checkImageBuildResponse` to take one argument again? Or maybe remove the image build log section in the README.md for now

internal/test/container.go
```diff
@@ -119,24 +135,49 @@ func runTest(ctx context.Context, logger *log.Logger, w io.Writer, dockerDaemon
 		return fmt.Errorf("failed to parse environment: %w", err)
 	}
 
+	if configuration.BroadcomProxy {
+		envMap["ARTIFACTORY_USERNAME"] = configuration.ArtifactoryUsername
+		envMap["ARTIFACTORY_PASSWORD"] = configuration.ArtifactoryPassword
+	}
+
 	artifactoryUsername := envMap["ARTIFACTORY_USERNAME"]
```
notrepo05: Nitpick: If the proxy is ever _not_ needed I'd recommend also supporting reading the ARTIFACTORY_USERNAME/PASSWORD via the OS environment too. If it isn't passed then these secrets are only configured from `-e`

notrepo05:
Works well and has a simple user experience. Thank you!

internal/commands/test_tile_test.go
```diff
@@ -226,29 +232,55 @@ var _ = Describe("kiln test", func() {
...
+					Expect(os.Unsetenv("ARTIFACTORY_USERNAME")).To(Succeed())
+				}
+				if hasP {
+					Expect(os.Setenv("ARTIFACTORY_PASSWORD", savedP)).To(Succeed())
+				} else {
+					Expect(os.Unsetenv("ARTIFACTORY_PASSWORD")).To(Succeed())
+				}
+			})
+			Expect(os.Unsetenv("ARTIFACTORY_USERNAME")).To(Succeed())
+			Expect(os.Unsetenv("ARTIFACTORY_PASSWORD")).To(Succeed())
```
notrepo05: Does the BeforeEach() set these environment variables before every test is executed? If so, we might be able to get away with

```suggestion
		    t := GinkgoT()
		    t.Setenv("ARTIFACTORY_USERNAME", "")
	        t.Setenv("ARTIFACTORY_PASSWORD", "")
```

internal/test/Dockerfile
notrepo05, line 48: The integration test will fail in the `migration` test unless we account for proxying npm. Verified the following works (docker image is local only, never pushed)

```suggestion
RUN which ops-manifest

RUN printf '%s\n%s\n' \
    'registry=https://${ARTIFACTORY_USERNAME}:${ARTIFACTORY_PASSWORD}@usw1.packages.broadcom.com/artifactory/api/npm/tis-npm-virtual/' \
    'always-auth=true' > /root/.npmrc
```

```
Ginkgo ran 1 suite in 12.107268128s
Test Suite Passed
[10:23:38] Completed: Manifest Tests

[10:22:41] ✓ Migration Tests Passed
[10:23:26] ✓ Stability Tests Passed
[10:23:38] ✓ Manifest Tests Passed
--- PASS: TestDockerIntegration (97.42s)
    --- PASS: TestDockerIntegration/the_test_succeeds (97.26s)
PASS
ok  	github.com/pivotal-cf/kiln/internal/test	97.765s
```

otherwise 

```
[10:16:46] ✗ Migration Tests Failed
[10:17:28] ✓ Stability Tests Passed
[10:17:41] ✓ Manifest Tests Passed
    integration_test.go:61:
        	Error Trace:	/Users/nr028083/workspace/kiln/internal/test/integration_test.go:61
        	Error:      	Received unexpected error:
        	            	test failed with exit code 1
        	Test:       	TestDockerIntegration/the_test_succeeds
--- FAIL: TestDockerIntegration (138.95s)
    --- FAIL: TestDockerIntegration/the_test_succeeds (138.83s)
FAIL
FAIL	github.com/pivotal-cf/kiln/internal/test	139.319s
```

notrepo05 (changes requested):
This works as expected running against real tiles. The fast fail is a nice ux improvement and the summary is easier to read!

I noticed that the integration test, due to network conditions, is failing. I would suggest making the verbosity more configurable so the errors related to that are visible:

```diff
diff --git a/internal/test/integration_test.go b/internal/test/integration_test.go
index c7bbb4f3..f410461b 100644
--- a/internal/test/integration_test.go
+++ b/internal/test/integration_test.go
@@ -50,6 +50,7 @@ func TestDockerIntegration(t *testing.T) {
 			AbsoluteTileDirectory: tmpDir,
 			RunAll:                true,
 			Environment:           []string{"ARTIFACTORY_USERNAME=" + artifactoryUsername, "ARTIFACTORY_PASSWORD=" + artifactoryPassword},
+			Verbose:               testing.Verbose(),
 		}
 		out := io.Discard
 		if testing.Verbose() {
```

But after applying the review comments below the test suite passes. 

I also left a nit on a test setup. 

Thank you!

internal/test/container.go
rizwanreza: We've assumed the proxy is always needed, but the env vars are also read from the shell.

internal/commands/test_tile_test.go
rizwanreza: Fixed. I like this!!

notrepo05 (approved):

PR #663, opened 2026-06-16
feat(carvel): add BOSH link support via job-spec-overlay and values-overlay sidecars

drich10:
## Summary

Introduce two per-packageinstall sidecar conventions that kiln auto-detects in `packageinstalls/`:

| Sidecar | Purpose | When it runs |
|---|---|---|
| `<entry>.job-spec-overlay.yml` | Declares additional BOSH link consumptions; kiln appends them to the generated `registry-data` job.MF | `kiln carvel bake` time |
| `<entry>.values-overlay.erb` | ERB fragment injected before `YAML.dump(values)`; mutates the values hash via BOSH link objects | BOSH deploy time (ERB) |

Missing sidecar files are silently skipped (no-op). Both sidecars are co-located with the packageinstall YAML they extend.

## Motivation

Carvel tiles need to consume BOSH links from co-deployed products (e.g. the `binding_cache` link provided by CF's `loggr-syslog-binding-cache` job) to inject environment-specific runtime values at BOSH deploy time. Previously only the hardcoded `cluster-info` link was available in the generated job spec. The binding-cache has no BOSH DNS alias, so its address can only be discovered via a BOSH link.

## Approach — co-located sidecars (mirrors the values-overlay.erb pattern)

Tile authors declare BOSH link consumptions in a `<name>.job-spec-overlay.yml` sidecar alongside their packageinstall YAML. kiln reads all such files and aggregates their `consumes:` entries into the generated `registry-data` job spec. The schema mirrors the BOSH job spec `consumes:` block directly.

```
packageinstalls/
  tnz-ear-runtime-package-install.yml
  tnz-ear-runtime-package-install.job-spec-overlay.yml   ← NEW: BOSH link declarations
  tnz-ear-runtime-package-install.values-overlay.erb     ← NEW: values mutations at deploy time
```

Example `job-spec-overlay.yml`:
```yaml
consumes:
- name: binding_cache
  type: binding_cache
  optional: false
```

Example `values-overlay.erb`:
```erb
<% addr = link("binding_cache").instances.first.address %>
<% values["syslog_agent"]["cache"]["url"] = "https://#{addr}:9000" %>
```

**Kilnfile is not modified.** It stays focused on dependency management (release sources, releases, stemcell). The `BOSHLinkConsumer`/`BOSHLinks` types and `BoshLinks` field from the initial commit have been removed.

## Backward Compatibility

- Tiles with no `*.job-spec-overlay.yml` files produce identical output to before (no-op)
- Tiles with no `*.values-overlay.erb` files produce identical output to before (no-op)
- All existing `cluster-info` link behaviour is preserved
- Kilnfile schema unchanged

## Test Plan

- [x] `jobSpecOverlay` — parses a `consumes:` list, handles empty list, handles missing key
- [x] `buildRegistryDataSpec` — includes user-declared links after cluster-info
- [x] `buildRegistryDataSpec` — no additional links when no sidecars present
- [x] `buildRegistryDataSpec` — optional flag rendered correctly
- [x] `generateManifestTemplate` with overlay — overlay content present and before `YAML.dump`
- [x] `generateManifestTemplate` with empty overlay — still produces valid template
- [x] Full suite: `go test ./...` — 0 failures (Docker integration test skipped due to no daemon)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/carvel/baker.go
+++ b/internal/carvel/baker.go
@@ -292,6 +292,12 @@ files:
 		}
 	}
 
+	// Read bosh_links from Kilnfile (optional; missing file or missing field is fine).
+	var kilnfileData cargo.Kilnfile
+	if kfData, err := os.ReadFile(path.Join(b.source, "Kilnfile")); err == nil {
+		_ = yaml.Unmarshal(kfData, &kilnfileData)
+	}
+
 	registryDataTemplates := ""
 	registryDataProperties := ""
 
@@ -343,7 +349,13 @@ files:
 			return err
 		}
 
-		manifestTemplate := generateManifestTemplate(entry)
+		// Read optional values-overlay ERB file alongside the packageinstall YAML.
+		overlayContent := ""
+		if overlayData, overlayErr := os.ReadFile(path.Join(b.source, "packageinstalls", entry+".values-overlay.erb")); overlayErr == nil {
+			overlayContent = string(overlayData)
+		}
+
+		manifestTemplate := generateManifestTemplate(entry, overlayContent)
 
 		err = os.WriteFile(
 			path.Join(dirName, "jobs", "registry-data", "templates", "packageinstalls", entry+".yml.erb"),
@@ -355,18 +367,7 @@ files:
 		}
 	}
 
-	registryDataSpec := `---
-name: registry-data
-templates:
-` + registryDataTemplates +
-		`packages:
-- registry-data
-consumes:
-- name: cluster
-  type: cluster-info
-  optional: true
-properties:
-` + registryDataProperties
+	registryDataSpec := buildRegistryDataSpec(registryDataTemplates, registryDataProperties, kilnfileData.BoshLinks.Consumes)
 
 	err = os.WriteFile(path.Join(dirName, "jobs", "registry-data", "spec"), []byte(registryDataSpec), 0644)
 	if err != nil {
@@ -376,7 +377,37 @@ properties:
 	return nil
 }
 
-func generateManifestTemplate(entry string) string {
+// buildRegistryDataSpec constructs the job.MF content for the registry-data BOSH job.
+// It always includes the hardcoded cluster-info link and appends any additional links
+// declared via bosh_links.consumes in the Kilnfile.
+func buildRegistryDataSpec(templates, properties string, additionalLinks []cargo.BOSHLinkConsumer) string {
+	extraLinks := ""
+	for _, link := range additionalLinks {
+		optional := "false"
+		if link.Optional {
+			optional = "true"
+		}
+		extraLinks += fmt.Sprintf("- name: %s\n  type: %s\n  optional: %s\n", link.Name, link.Type, optional)
+	}
+	return `---
+name: registry-data
+templates:
+` + templates +
+		`packages:
+- registry-data
+consumes:
+- name: cluster
+  type: cluster-info
+  optional: true
+` + extraLinks +
+		`properties:
+` + properties
+}
+
+// generateManifestTemplate produces the ERB template for the registry-data BOSH job.
+// overlayContent is optional ERB code injected into the values manipulation block
+// before YAML.dump(values) is called, enabling BOSH link-based value overrides.
+func generateManifestTemplate(entry, overlayContent string) string {
 	return `---
 apiVersion: v1
 kind: ServiceAccount
@@ -423,6 +454,7 @@ stringData:
     values["context"]["namespace"] = link("cluster").p("content-namespace") rescue "default"
   end
 %>
+` + overlayContent + `
 <%= YAML.dump(values).split("\n").map { |line| "    " + line }.join("\n") %>
 ---
 apiVersion: packaging.carvel.dev/v1alpha1
--- a/internal/carvel/baker.go
+++ b/internal/carvel/baker.go
@@ -255,6 +255,33 @@ func (b *baker) progress(message string) {
 	_, _ = fmt.Fprintln(b.progressWriter, message)
 }
 
+// deduplicateConsumes removes duplicate BOSH link consumer entries by name.
+// Identical duplicates are dropped silently. If two entries share a name but
+// differ in type or optional, the first is kept and a WARNING is emitted —
+// BOSH rejects duplicate link names in job.MF, so the second is always ignored.
+func (b *baker) deduplicateConsumes(consumes []boshLinkConsumer) []boshLinkConsumer {
+	seen := make(map[string]boshLinkConsumer)
+	var deduped []boshLinkConsumer
+	for _, c := range consumes {
+		existing, ok := seen[c.Name]
+		if !ok {
+			seen[c.Name] = c
+			deduped = append(deduped, c)
+			continue
+		}
+		if existing != c {
+			b.progress(fmt.Sprintf(
+				"WARNING: duplicate BOSH link consumer name %q found across packageinstalls.\n"+
+					"  Keeping:  {type: %s, optional: %v}\n"+
+					"  Ignoring: {type: %s, optional: %v}\n"+
+					"  Ensure all packageinstalls agree on the link definition.",
+				c.Name, existing.Type, existing.Optional, c.Type, c.Optional,
+			))
+		}
+	}
+	return deduped
+}
+
 // boshLinkConsumer declares a BOSH link the registry-data job should consume.
 // Populated from per-packageinstall *.job-spec-overlay.yml sidecar files.
 type boshLinkConsumer struct {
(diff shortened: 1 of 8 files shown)


rizwanreza:
I have a concern on bosh_links placement: should this live in Kilnfile?

Kilnfile is a dependency manifest. It tracks which BOSH releases to fetch and where from (think Gemfile). bosh_links is deploy-time BOSH job config. Those are different concerns, and mixing them here will confuse anyone reading the Kilnfile expecting only release sourcing info.

Three alternatives worth discussing:

**Option A — Sidecar file (matches this PR's own pattern)**

This PR already introduces *.values-overlay.erb as a co-located sidecar kiln auto-detects. Same idea works here: a registry-data-consumes.yml kiln reads and merges into the generated job spec. Keeps things co-located and Kilnfile clean.

**Option B — base.yml with a kiln extension stanza**
base.yml is already the tile's primary config, already processed by kiln. A carvel: block kiln reads and strips before handing off to OpsManager. Everything tile-level in one place.

**Option C — Dedicated carvel.yml**
Standalone file for carvel-specific build config. Cleanest separation, but adds a new convention.

Option A feels most natural given what's already in this PR. Open to other takes.

drich10:
@rizwanreza I like option A - any strong idea of how you'd describe the "overlay" section for the job spec vs the link definition section? I'll update the design/pr later today with this in mind

drich10:
@rizwanreza Updated — went with a variant of your Option A that mirrors the `*.values-overlay.erb` pattern more closely.

Instead of `registry-data-consumes.yml` at the tile root, the sidecar is named `<entry>.job-spec-overlay.yml` and lives alongside the packageinstall YAML in `packageinstalls/`. This keeps both sidecars co-located with the entry they extend and uses a consistent `*-overlay.*` naming convention:

| Sidecar | Scope | Content |
|---|---|---|
| `<entry>.job-spec-overlay.yml` | Build time | `consumes:` list for the BOSH job spec |
| `<entry>.values-overlay.erb` | Deploy time | ERB mutations using BOSH link objects |

`bosh_links` has been removed from Kilnfile entirely — the `BOSHLinkConsumer`/`BOSHLinks` types are gone from `pkg/cargo` and Kilnfile is back to being a pure dependency manifest.

internal/carvel/baker.go
```diff
@@ -362,7 +378,22 @@ files:
...
+		overlayContent := ""
+		if overlayData, overlayErr := os.ReadFile(path.Join(b.source, "packageinstalls", entry+".values-overlay.erb")); overlayErr == nil {
+			overlayContent = string(overlayData)
+		}
+
+		// Read optional job-spec-overlay sidecar to discover additional BOSH link consumptions.
+		jobSpecOverlayPath := path.Join(b.source, "packageinstalls", entry+".job-spec-overlay.yml")
+		if overlayData, overlayErr := os.ReadFile(jobSpecOverlayPath); overlayErr == nil {
+			var overlay jobSpecOverlay
+			if parseErr := yaml.Unmarshal(overlayData, &overlay); parseErr == nil {
```
davewalter: 🤖 AI Review
---
🔴 **Must Fix**

When the sidecar file exists but contains malformed YAML, the parse error is silently swallowed. The tile is then built without the intended BOSH link declarations and will fail silently at deploy time — the author gets no feedback that their overlay was ignored.

Consider returning the error:
```go
if parseErr := yaml.Unmarshal(overlayData, &overlay); parseErr != nil {
    return fmt.Errorf("parsing %s: %w", jobSpecOverlayPath, parseErr)
}
```
If silent-skip is truly desired (e.g., for forward compatibility), at minimum log a warning via `b.log()`.

internal/carvel/baker.go
```diff
@@ -362,7 +378,22 @@ files:
 			return err
 		}
 
-		manifestTemplate := generateManifestTemplate(entry)
+		// Read optional values-overlay ERB file alongside the packageinstall YAML.
+		overlayContent := ""
+		if overlayData, overlayErr := os.ReadFile(path.Join(b.source, "packageinstalls", entry+".values-overlay.erb")); overlayErr == nil {
```
davewalter: 🤖 AI Review
---
🟡 **Should Fix**

Both sidecar reads use `if ...; overlayErr == nil` which treats **all** OS errors identically — a permission-denied or a directory-where-file-expected is silently skipped just like a not-found. The intent is to silently skip missing files, but unexpected errors should surface.

Swap to an explicit not-found check:
```go
if overlayData, overlayErr := os.ReadFile(...); overlayErr == nil {
    overlayContent = string(overlayData)
} else if !os.IsNotExist(overlayErr) {
    return overlayErr  // or b.log a warning
}
```
The same pattern applies to the job-spec-overlay read on the next block.

internal/carvel/baker.go
```diff
@@ -362,7 +378,22 @@ files:
...
+			overlayContent = string(overlayData)
+		}
+
+		// Read optional job-spec-overlay sidecar to discover additional BOSH link consumptions.
+		jobSpecOverlayPath := path.Join(b.source, "packageinstalls", entry+".job-spec-overlay.yml")
+		if overlayData, overlayErr := os.ReadFile(jobSpecOverlayPath); overlayErr == nil {
+			var overlay jobSpecOverlay
+			if parseErr := yaml.Unmarshal(overlayData, &overlay); parseErr == nil {
+				allConsumes = append(allConsumes, overlay.Consumes...)
```
davewalter: 🤖 AI Review
---
🟡 **Should Fix**

`allConsumes` accumulates entries from every packageinstall with no deduplication. If two packageinstalls both declare a link with the same `name` (even identically defined), the generated `job.MF` will contain duplicate `consumes:` entries. BOSH rejects specs with duplicate link names.

Add a deduplication step before passing to `buildRegistryDataSpec`, keyed on `link.Name`. Alternatively, detect the duplicate here and return an error that tells the author which files conflict.

internal/carvel/baker.go
```diff
@@ -374,28 +405,47 @@ files:
...
+// It always includes the hardcoded cluster-info link and appends any additional links
+// collected from *.job-spec-overlay.yml sidecars in the packageinstalls/ directory.
+func buildRegistryDataSpec(templates, properties string, additionalLinks []boshLinkConsumer) string {
+	extraLinks := ""
+	for _, link := range additionalLinks {
+		optional := "false"
+		if link.Optional {
+			optional = "true"
+		}
+		extraLinks += fmt.Sprintf("- name: %s\n  type: %s\n  optional: %s\n", link.Name, link.Type, optional)
```
davewalter: 🤖 AI Review
---
🟡 **Should Fix**

`link.Name` and `link.Type` are interpolated directly into a raw YAML string via `fmt.Sprintf`. A value containing a newline, colon, or other YAML-special character would silently produce a malformed (or structurally different) `job.MF`. Since `boshLinkConsumer` is already a typed struct, prefer marshalling the `consumes:` block with the yaml library:

```go
extraBytes, err := yaml.Marshal(map[string]any{"consumes": additionalLinks})
if err != nil {
    return "", err
}
// then splice into the spec string
```
(This would require changing the return type of `buildRegistryDataSpec` to `(string, error)`.)

internal/carvel/baker_test.go
```diff
@@ -103,6 +103,82 @@ var _ = Describe("Carvel Baker", func() {
...
+
+		It("handles a missing consumes key without error", func() {
+			var overlay jobSpecOverlay
+			err := yaml.Unmarshal([]byte("{}"), &overlay)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(overlay.Consumes).To(BeNil())
+		})
 	})
 
 	Context("Bake", func() {
```
davewalter, line 285: 🤖 AI Review
---
🟡 **Should Fix**

The integration test (`Bake` context) exercises the full pipeline via `testdata/sample-tile` but that test data doesn't include any `.job-spec-overlay.yml` or `.values-overlay.erb` sidecars. The unit tests for the helper functions are good, but there's no end-to-end coverage of the new sidecar loading path through `generateBoshReleaseDir`. Consider adding a variant of `testdata/sample-tile` (or a fixture file) that includes at least one `job-spec-overlay.yml` and verifying it appears in the generated `spec` file.

internal/carvel/baker.go
```diff
@@ -374,18 +416,19 @@ files:
...
-properties:
-` + registryDataProperties
+	seen := make(map[string]struct{})
+	var deduped []boshLinkConsumer
+	for _, c := range allConsumes {
+		if _, ok := seen[c.Name]; !ok {
+			seen[c.Name] = struct{}{}
+			deduped = append(deduped, c)
+		}
+	}
```
rizwanreza: The deduplication keeps the first occurrence and discards the rest. If packageinstall A declares binding_cache with type: binding_cache and packageinstall B declares binding_cache with type: binding-cache-v2, B's type is silently dropped. The generated BOSH spec will use A's type with no indication that B's declaration was ignored. This would be a confusing bug to track down at deploy time.

rizwanreza (commented):
Just a minor thing but otherwise look great!

internal/carvel/baker.go
drich10: good call - we log this now in the output since I think its a _reasonable_ decision but puts the information in the tile author's view during build. WDYT?

PR #664, opened 2026-06-18
feat(carvel): support structured BOSH variable declarations in base.yml

drich10:
## Summary

- Changes `Variables []string` → `Variables []proofing.Variable` in `internal/carvel/models/metadata.go` and `metadata_out.go`
- Adds `validateVariables()` to `baker.go` — returns a clear error if a variable entry is missing `name` or `type`
- Covers with 4 new Ginkgo unit specs + updates `testdata/sample-tile/base.yml` for integration round-trip coverage

## Motivation

Carvel tile authors need to declare BOSH CredHub variable definitions (e.g., certificate CA hierarchies) inline in `base.yml`. The previous `[]string` type rejected structured map entries with a YAML unmarshal error:

```
yaml: unmarshal errors:
  line 34: cannot unmarshal !!map into string
```

Discovered while building the EAR-K8s tile's Diego Instance Identity CA support. The EAR-K8s tile mirrors the TAS IST tile's pattern of declaring `diego-instance-identity-intermediate-ca-2-7` as a BOSH CredHub certificate variable so BOSH manages the CA lifecycle automatically.

## Why `proofing.Variable`?

`pkg/proofing` already models this shape (`name`, `type`, `options any`). Reusing it avoids duplicating the struct. The import direction (`internal/carvel` → `pkg/proofing`) is valid — no circular dependency.

## Test Plan

- [x] `go test ./internal/carvel/... --ginkgo.focus="validateVariables"` — 4 new unit specs pass
- [x] `go test ./internal/carvel/...` — 36/36 specs pass, including integration round-trip for structured variables
- [x] `go build ./internal/carvel/...` — clean build

## Related

TNZ-112157

Made with [Cursor](https://cursor.com)


internal/carvel/baker_test.go
```diff
@@ -169,7 +170,9 @@ var _ = Describe("Carvel Baker", func() {
 					Expect(outMeta.Serial).To(BeFalse())
 					Expect(outMeta.PropertyBlueprints).To(HaveLen(2))
 					Expect(outMeta.FormTypes).To(HaveLen(1))
-					Expect(outMeta.Variables).To(BeEmpty())
+					Expect(outMeta.Variables).To(HaveLen(1))
+				Expect(outMeta.Variables[0].Name).To(Equal("sample-tile-ca"))
+				Expect(outMeta.Variables[0].Type).To(Equal("certificate"))
```
davewalter: 🤖 AI Review
---
🟡 **Should Fix**

These two new `Expect` lines are indented with 4 tabs while every surrounding assertion uses 5 tabs. This will fail `gofmt` checks if enforced in CI.

```go
// Should match the surrounding 5-tab style:
					Expect(outMeta.Variables[0].Name).To(Equal("sample-tile-ca"))
					Expect(outMeta.Variables[0].Type).To(Equal("certificate"))
```

davewalter: 🤖 AI Review
---
🟡 **Should Fix**

The PR title says "round-trip coverage" but the test only asserts `Name` and `Type` — the `options` block (`common_name`, `is_ca`, `duration`) is not verified. Since `Options` is typed as `any`, mixed-type round-trip fidelity (bool, int, string) is not guaranteed without an explicit assertion.

Suggested addition:
```go
Expect(outMeta.Variables[0].Options).To(HaveKeyWithValue("common_name", "Sample Tile CA"))
Expect(outMeta.Variables[0].Options).To(HaveKeyWithValue("is_ca", true))
```

internal/carvel/baker.go
```diff
@@ -248,6 +255,19 @@ func (b *baker) progress(message string) {
...
 }
 
+// validateVariables checks that each variable declaration has the required name and type fields.
+func validateVariables(vars []proofing.Variable) error {
+	for i, v := range vars {
+		if v.Name == "" {
+			return fmt.Errorf("variables[%d]: missing required field 'name'", i)
+		}
+		if v.Type == "" {
+			return fmt.Errorf("variables[%d] (%q): missing required field 'type'", i, v.Name)
```
davewalter: 🤖 AI Review
---
🟢 **Suggestion**

`validateVariables` returns on the first invalid entry, so an author with multiple malformed variables must fix them one at a time. If the project is on Go 1.20+, `errors.Join` makes it trivial to report all errors at once:

```go
func validateVariables(vars []proofing.Variable) error {
	var errs []error
	for i, v := range vars {
		if v.Name == "" {
			errs = append(errs, fmt.Errorf("variables[%d]: missing required field 'name'", i))
		} else if v.Type == "" {
			errs = append(errs, fmt.Errorf("variables[%d] (%q): missing required field 'type'", i, v.Name))
		}
	}
	return errors.Join(errs...)
}
```

davewalter (approved):

PR #669, opened 2026-07-27
feat: Updated kiln carvel command to include -pkg suffix to bosh release names

ram-pivot:
ai-assisted=yes
JIRA-ID: 119962

### DIFF
--- a/internal/carvel/baker.go
+++ b/internal/carvel/baker.go
@@ -91,6 +91,9 @@ func (b *baker) Bake(source string) error {
 	if err != nil {
 		return err
 	}
+	if b.metadata.Name == "" {
+		return errors.New("missing required field 'name' in tile metadata (base.yml)")
+	}
 	if err := validateVariables(b.metadata.Variables); err != nil {
 		return err
 	}
@@ -142,6 +145,9 @@ func (b *baker) BakeFromLockfile(source string, releaseLock cargo.BOSHReleaseTar
 	if err != nil {
 		return err
 	}
+	if b.metadata.Name == "" {
+		return errors.New("missing required field 'name' in tile metadata (base.yml)")
+	}
 	if err := validateVariables(b.metadata.Variables); err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


internal/carvel/baker.go
```diff
@@ -151,8 +152,8 @@ func (b *baker) BakeFromLockfile(source string, releaseLock cargo.BOSHReleaseTar
 	}
 	b.progress(fmt.Sprintf("Tile: %s version %s (metadata_version %s)", b.metadata.Name, ver, b.metadata.MetadataVersion))
 
-	if releaseLock.Name != b.metadata.Name {
-		return fmt.Errorf("lockfile release name %q does not match tile name %q", releaseLock.Name, b.metadata.Name)
+	if releaseLock.Name != b.GetBoshReleaseName() {
+		return fmt.Errorf("lockfile release name %q does not match tile name %q", releaseLock.Name, b.GetBoshReleaseName())
```
notrepo05: It might be clearer if we update this error message to also mention the new suffix. 
```suggestion
		return fmt.Errorf("lockfile release name %q does not match tile-derived name %q", releaseLock.Name, b.GetBoshReleaseName())
```

internal/carvel/baker.go
```diff
@@ -221,6 +222,13 @@ func (b *baker) GetName() string {
 	return b.metadata.Name
 }
 
+func (b *baker) GetBoshReleaseName() string {
+	if b.metadata.Name == "" {
```
notrepo05, line 232: If it's possible, it could be useful to verify metadata.Name isn't empty earlier.

notrepo05 (approved):
I have a couple of minor nits, but overall this looks good. Thanks!

It might be worth considering having carvel-rebake enforce that the kiln version is the same one as recorded in the bake record, since it'll make this breaking change easier mitigate / more intuitive.

internal/carvel/baker.go
ram-pivot: Done.

internal/carvel/baker.go
ram-pivot: Done.

ram-pivot:
> I have a couple of minor nits, but overall this looks good. Thanks!
> 
> It might be worth considering having carvel-rebake enforce that the kiln version is the same one as recorded in the bake record, since it'll make this breaking change easier mitigate / more intuitive.

Thanks for the review. Interestingly kiln carvel publish does not write the kiln version. I fixed that and also added the kiln version check

https://github.gwd.broadcom.net/TNZ/tac-helm-cf-tile/blob/main/applications/strimzi/tile/bake_records/10.4.3.json

notrepo05 (approved):

rizwanreza (approved):

PR #672, opened 2026-07-31
Implement remote fetch, carvel publish, and hook synthesis for carvel tiles

ichandrabhatta:
### Summary
Extends `kiln carvel bake` so Carvel tiles can declare third-party BOSH releases (`additional_releases`) fetched from real remote sources via `Kilnfile`/`Kilnfile.lock` — instead of requiring pre-staged local tarballs — and auto-synthesizes pre/post-install hook adapter jobs directly into the tile's own release. Also brings `carvel upload`/`publish`/`rebake` up to the same standard as classic `kiln bake`: a cached, checksummed release in Artifactory, a reproducible `.pivotal` built from it, and a bake record for `rebake` to verify against later.

### Key Features

1. **Remote fetch for `additional_releases`**
   - `additional_releases:` entries are now resolved via `Kilnfile`/`Kilnfile.lock`'s `release_sources:` (`fetchAdditionalReleases`), reusing the same `component.NewReleaseSourceRepo`/`DownloadRelease` primitive classic `kiln bake` already uses — not a local `tarball_path:` anymore.
   - New `carvel bake` flags: `--releases-directory` (default `releases`), `--skip-fetch` (use whatever's already staged there), `--from-lockfile` (explicit opt-in to bake the tile's *own* release from a cached copy instead of regenerating it — see bug fix below for why this is now explicit).
   - Local SHA1 cache-hit check avoids re-downloading a release that's already present and matches the lock.

2. **Auto-synthesis of pre/post-install hooks**
   - `pre_install_hooks:`/`post_install_hooks:` in `base.yml` get a BOSH job auto-generated into the tile's own release (auto-prefixed with the tile's name to avoid collisions), depositing a single script at `bin/hooks/{pre,post}-install` that `exec`s the declared command.
   - Eliminates the need for tile authors to maintain separate "hook adapter" BOSH release repositories.

3. **`carvel upload` / `publish` / `rebake` + bake records**
   - `carvel upload` builds the tile's own release, uploads it to Artifactory, and records it in `Kilnfile.lock`.
   - `carvel publish` downloads that cached release (by name, not by list position) and bakes the final `.pivotal`; `--final` writes a `bake_records/<version>.json` (git SHA + tile SHA256 + version).
   - `carvel rebake` reproduces a past release from its bake record and verifies the checksum matches.
   - `Kilnfile.lock`'s upsert logic (`writeStandardKilnfileLock`/`UpdateBOSHReleaseTarballLockWithName`) now merges by release name instead of overwriting the whole file — the tile's own release and its `additional_releases` entries coexist in one lockfile.

### Bug Fixes / Robustness (found and fixed during review)

- **Critical**: `carvel bake`'s mode dispatch used to switch to "bake from cached copy" purely because `Kilnfile.lock` existed on disk — silently skipping hook/registry-data synthesis. Now gated behind explicit `--from-lockfile`.
- **Critical**: `downloadCarvelRelease`/`generateRuntimeConfigs`'s cache/lock lookups used to assume "the tile's own release" was always `Releases[0]` — broke the moment `additional_releases` shared the same lockfile. Now resolved by name everywhere.
- A malformed `Kilnfile.lock` used to be silently swallowed (falls back to an empty lock, surfaces a confusing unrelated error later); now hard-fails immediately, while still tolerating a genuinely absent lockfile.
- `additional_releases` cached-file lookup assumed every release source names its downloaded tarball `<name>-<version>.tgz`; `s3`/`artifactory` sources actually write `filepath.Base(RemotePath)` (e.g. `smoke-tests-4.12.14-ubuntu-jammy-1.1298.tgz`), which broke `--skip-fetch` and the cache-hit check for those sources. Now resolves the real source type before deciding the filename convention.
- Hook validation (missing name/command) was only enforced on the `bake` path; `publish`/`rebake` (via `BakeFromLockfile`) skipped it. Now enforced on both.
- Restored a dropped diagnostic in `update-release` (the "-release/-boshrelease suffix" hint) and fixed a stale test that had silently stopped asserting on it.
- De-duplicated the hook job-name auto-prefix logic (previously implemented twice, at risk of drifting out of sync between job generation and runtime-config reference).
- Fixed several stale/self-contradicting code comments.
- Fixed 4 test call sites still using an old function signature, which had been failing `go vet`/`go test` outright.

### Known residual gap (documented, not fixed here)
`publish`/`rebake` rebuild the runtime-config's hook/job references from whatever `base.yml` currently says, independent of what's actually inside the cached release tarball built by `upload`. If hooks change between `upload` and `publish` without re-running `upload`, the two can drift. Currently an operator contract ("re-run `upload` if you change hooks"), not enforced in code.

### Technical Details
* **`internal/carvel/models/metadata.go`**: `AdditionalReleases`, `PreInstallHooks`, `PostInstallHooks` fields on `Metadata`; `AdditionalRelease`/`AdditionalJob`/`HookDeclaration` structs.
* **`internal/carvel/baker.go`**: `fetchAdditionalReleases`, `hookJobName` (shared prefix helper), `additionalReleaseLocalFilename` (source-type-aware cache filename resolution), hook synthesis in `generateBoshReleaseDir`, hook/additional-release wiring in `generateRuntimeConfigs`, `BakeOptions` (`SkipFetch`, `ReleasesDirectory`) threaded through `Bake`/`BakeFromLockfile`.
* **`internal/commands/carvel_bake.go`**: `--releases-directory`/`--skip-fetch`/`--from-lockfile` flags; Kilnfile/Kilnfile.lock load-error handling that distinguishes "missing" from "malformed".
* **`internal/commands/carvel_helpers.go`**: `downloadCarvelRelease`/`writeStandardKilnfileLock` resolve by release name, not list position; upsert instead of overwrite.
* **`internal/commands/carvel_upload.go` / `carvel_publish.go` / `carvel_rebake.go`**: updated to the new `BakeOptions`/name-based lookup plumbing.
* **`internal/commands/update_release.go`**, **`pkg/cargo/kilnfile.go`**: minor fixes described above.
* **Tests**: `internal/carvel/baker_test.go`, `internal/commands/carvel_bake_test.go`, `internal/commands/carvel_helpers_test.go`, `pkg/cargo/kilnfile_test.go` — coverage for hook synthesis, remote-fetch/cache-hit behavior (including the source-type filename mismatch), lockfile-preservation, and the restored diagnostics.

### DIFF
--- a/internal/acceptance/carvel/carvel_workflow_test.go
+++ b/internal/acceptance/carvel/carvel_workflow_test.go
@@ -24,11 +24,6 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// mockArtifactory is a test HTTP server that faithfully simulates Artifactory's
-// upload (PUT) and download (GET) behaviour, including Basic Auth verification.
-// Upload stores the tarball bytes keyed by request path; download serves them
-// back.  The /artifactory prefix that the real download client prepends is
-// handled transparently.
 type mockArtifactory struct {
 	mu       sync.Mutex
 	blobs    map[string][]byte
@@ -57,7 +52,6 @@ func (m *mockArtifactory) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Normalize path: strip the /artifactory prefix the download client adds
 	key := r.URL.Path
 	key = strings.TrimPrefix(key, "/artifactory")
 
@@ -114,23 +108,6 @@ func (m *mockArtifactory) GetCount() int {
 	return m.getCount
 }
 
-// ---------------------------------------------------------------------------
-// End-to-end acceptance test for the full Carvel tile developer workflow.
-//
-// This test exercises every `kiln carvel` subcommand in sequence, chaining
-// their outputs exactly like a real developer and CI pipeline would:
-//
-//   Step 1  kiln carvel bake          (local bake, no Kilnfile.lock)
-//   Step 2  kiln carvel upload        (creates BOSH release, uploads, writes lock)
-//           git add + commit          (Kilnfile.lock)
-//   Step 3  kiln carvel bake          (CI-style: downloads from lock)
-//   Step 4  kiln carvel publish --final (downloads, bakes, writes bake record)
-//           git add + commit          (bake_records/)
-//   Step 5  kiln carvel rebake        (re-bakes from record, verifies checksum)
-//
-// A mock Artifactory server stores the actual uploaded tarball and serves it
-// back on download, proving the full round-trip.
-// ---------------------------------------------------------------------------
 var _ = Describe("carvel full workflow", Ordered, func() {
 	const (
 		sampleTileFixture = "fixtures/sample-tile"
@@ -232,9 +209,6 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 		_ = os.RemoveAll(tmpDir)
 	})
 
-	// -----------------------------------------------------------------------
-	// Step 1: Local bake (no Kilnfile.lock, no Artifactory interaction)
-	// -----------------------------------------------------------------------
 	It("Step 1: bakes a tile locally without Kilnfile.lock", func() {
 		outputFile := filepath.Join(tmpDir, "step1.pivotal")
 
@@ -259,9 +233,6 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 		Expect(os.IsNotExist(err)).To(BeTrue(), "local bake must not create Kilnfile.lock")
 	})
 
-	// -----------------------------------------------------------------------
-	// Step 2: Upload (creates BOSH release, uploads to Artifactory, writes lock)
-	// -----------------------------------------------------------------------
 	It("Step 2: uploads the BOSH release to Artifactory and writes Kilnfile.lock", func() {
 		cmd := exec.Command(pathToMain,
 			append([]string{
@@ -298,9 +269,6 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 		gitInTile("commit", "-m", "add Kilnfile.lock from upload")
 	})
 
-	// -----------------------------------------------------------------------
-	// Step 3: CI-style bake (downloads cached BOSH release via Kilnfile.lock)
-	// -----------------------------------------------------------------------
 	It("Step 3: bakes a tile using Kilnfile.lock (CI path with Artifactory download)", func() {
 		outputFile := filepath.Join(tmpDir, "step3-ci.pivotal")
 
@@ -311,6 +279,7 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 				"carvel", "bake",
 				"--source-directory", inputPath,
 				"--output-file", outputFile,
+				"--from-lockfile",
 				"--verbose",
 			}, variableFlags()...)...,
 		)
@@ -324,9 +293,6 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 			"CI bake must download the cached BOSH release from Artifactory")
 	})
 
-	// -----------------------------------------------------------------------
-	// Step 4: Publish --final (downloads, bakes, creates bake record)
-	// -----------------------------------------------------------------------
 	var publishChecksum string
 
 	It("Step 4: publishes a final tile and writes a bake record", func() {
@@ -372,9 +338,6 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 		// Concourse resource checks out the commit from the record.
 	})
 
-	// -----------------------------------------------------------------------
-	// Step 5: Rebake from bake record (reproducibility verification)
-	// -----------------------------------------------------------------------
 	It("Step 5: re-bakes from the bake record with an identical checksum", func() {
 		outputFile := filepath.Join(tmpDir, "step5-rebake.pivotal")
 		recordPath := filepath.Join(inputPath, "bake_records", "0.1.1.json")
--- a/internal/carvel/baker.go
+++ b/internal/carvel/baker.go
@@ -1,6 +1,7 @@
 package carvel
 
 import (
+	"crypto/sha1"
 	"crypto/sha256"
 	"encoding/hex"
 	"errors"
@@ -16,6 +17,7 @@ import (
 	"strings"
 
 	"github.com/pivotal-cf/kiln/internal/carvel/models"
+	"github.com/pivotal-cf/kiln/internal/component"
 	"github.com/pivotal-cf/kiln/pkg/cargo"
 	"github.com/pivotal-cf/kiln/pkg/proofing"
 
@@ -23,12 +25,18 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type BakeOptions struct {
+	SkipFetch         bool
+	ReleasesDirectory string
+}
+
 // Baker transforms an imgpkg bundle and tile metadata into a BOSH release
 // and kiln-compatible tile structure that can be baked into a .pivotal file.
 type Baker interface {
-	Bake(source string) error
-	BakeFromLockfile(source string, releaseLock cargo.BOSHReleaseTarballLock, localTarball string) error
+	Bake(source string, kilnfile cargo.Kilnfile, kilnfileLock cargo.KilnfileLock, opts BakeOptions) error
+	BakeFromLockfile(source string, kilnfile cargo.Kilnfile, kilnfileLock cargo.KilnfileLock, releaseLock cargo.BOSHReleaseTarballLock, localTarball string, opts BakeOptions) error
 	KilnBake(destination string) error
+	ParseMetadata(source string) error
 	GetName() string
 	GetBoshReleaseName() string
 	// GetVersion returns the product version from base.yml or the version file.
@@ -76,7 +84,7 @@ func (b *baker) KilnBake(destination string) error {
 	return nil
 }
 
-func (b *baker) Bake(source string) error {
+func (b *baker) Bake(source string, kilnfile cargo.Kilnfile, kilnfileLock cargo.KilnfileLock, opts BakeOptions) error {
 	b.source = source
 	b.destination = path.Join(source, ".carvel-tile")
 
@@ -121,7 +129,7 @@ func (b *baker) Bake(source string) error {
 	}
 
 	b.progress("Generating tile layout in " + b.destination)
-	err = b.generateOutputTile()
+	err = b.generateOutputTile(kilnfile, kilnfileLock, opts)
 	if err != nil {
 		b.log(err.Error())
 		return err
@@ -130,7 +138,7 @@ func (b *baker) Bake(source string) error {
 	return nil
 }
 
-func (b *baker) BakeFromLockfile(source string, releaseLock cargo.BOSHReleaseTarballLock, localTarball string) error {
+func (b *baker) BakeFromLockfile(source string, kilnfile cargo.Kilnfile, kilnfileLock cargo.KilnfileLock, releaseLock cargo.BOSHReleaseTarballLock, localTarball string, opts BakeOptions) error {
 	b.source = source
 	b.destination = path.Join(source, ".carvel-tile")
 
@@ -210,6 +218,16 @@ func (b *baker) BakeFromLockfile(source string, releaseLock cargo.BOSHReleaseTar
 		return fmt.Errorf("failed to copy cached release tarball: %w", err)
 	}
 
+	// We also need to fetch any additional releases when baking from lockfile,
+	// otherwise the final .pivotal assembly will fail because it looks for them.
+	if len(b.metadata.AdditionalReleases) > 0 {
+		b.progress("  Fetching additional BOSH releases")
+		err = b.fetchAdditionalReleases(kilnfile, kilnfileLock, opts)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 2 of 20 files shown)


pkg/cargo/kilnfile.go
```diff
@@ -45,14 +45,15 @@ func (k KilnfileLock) FindBOSHReleaseWithName(name string) (BOSHReleaseTarballLo
 	return BOSHReleaseTarballLock{}, errors.New("not found")
 }
 
-func (k KilnfileLock) UpdateBOSHReleaseTarballLockWithName(name string, lock BOSHReleaseTarballLock) error {
+func (k *KilnfileLock) UpdateBOSHReleaseTarballLockWithName(name string, lock BOSHReleaseTarballLock) error {
 	for i, r := range k.Releases {
 		if r.Name == name {
 			k.Releases[i] = lock
 			return nil
```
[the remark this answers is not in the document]
ichandrabhatta: fixed

internal/commands/carvel_helpers.go
```diff
@@ -41,12 +41,24 @@ func findArtifactorySource(kilnfile cargo.Kilnfile) (cargo.ReleaseSourceConfig,
...
+	for _, r := range lock.Releases {
+		if r.Name == releaseName {
+			releaseLock = r
+			found = true
+			break
+		}
+	}
+	if !found {
+		return "", fmt.Errorf("release %q not found in Kilnfile.lock", releaseName)
+	}
```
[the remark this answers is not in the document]
ichandrabhatta: fixed

internal/commands/carvel_bake.go
```diff
@@ -54,36 +58,64 @@ func (c CarvelBake) Execute(args []string) error {
...
+	kf, kl, loadErr := c.Options.LoadKilnfiles(nil, nil)
+	if loadErr == nil {
+		kilnfile = kf
+		kilnfileLock = kl
+	} else if lockfilePresent {
+		return fmt.Errorf("failed to load Kilnfiles: %w", loadErr)
+	} else if c.Options.FromLockfile {
+		return fmt.Errorf("failed to load Kilnfiles (required for --from-lockfile): %w", loadErr)
+	} else {
+		c.outLogger.Printf("No Kilnfile/Kilnfile.lock found — proceeding without additional_releases support")
```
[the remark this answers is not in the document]
ichandrabhatta: fixed

internal/carvel/baker.go
```diff
@@ -723,10 +822,32 @@ func (b *baker) generateRuntimeConfigs() error {
...
+
+	for _, hook := range append(
+		append([]models.HookDeclaration{}, b.metadata.PreInstallHooks...),
+		b.metadata.PostInstallHooks...,
+	) {
+		if hook.Name == "" || hook.Command == "" {
+			return fmt.Errorf("hook declaration missing name or command")
+		}
+		addonJobs = append(addonJobs, models.Job{Name: b.hookJobName(hook.Name), Release: b.metadata.Name})
+	}
```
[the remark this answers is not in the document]
ichandrabhatta: fxied

internal/carvel/baker.go
```diff
@@ -455,6 +494,56 @@ files:
...
+
+			if err = os.MkdirAll(path.Join(dirName, "jobs", jobName, "templates"), 0755); err != nil {
+				return err
+			}
+			templateContent := fmt.Sprintf("#!/bin/bash\nset -euo pipefail\nexec %s\n", hook.Command)
+			err = os.WriteFile(
+				path.Join(dirName, "jobs", jobName, "templates", templateName),
+				[]byte(templateContent),
+				0644,
+			)
```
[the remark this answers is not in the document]
ichandrabhatta: fixed

internal/commands/carvel_upload.go
```diff
@@ -58,9 +60,15 @@ func (c CarvelUpload) Execute(args []string) error {
 	}
 
 	c.Options.Kilnfile = kilnfilePath
-	kilnfile, err := loadKilnfileOnly(c.Options.Standard)
+	kilnfile, kilnfileLock, err := c.Options.LoadKilnfiles(nil, nil)
 	if err != nil {
-		return fmt.Errorf("failed to load Kilnfile: %w", err)
+		if _, lockStatErr := os.Stat(c.Options.KilnfileLockPath()); lockStatErr == nil {
+			return fmt.Errorf("failed to load Kilnfile.lock: %w", err)
+		}
```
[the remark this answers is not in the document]
ichandrabhatta: fixed

internal/acceptance/carvel/carvel_workflow_test.go
```diff
@@ -311,6 +311,11 @@ var _ = Describe("carvel full workflow", Ordered, func() {
 				"carvel", "bake",
 				"--source-directory", inputPath,
 				"--output-file", outputFile,
+				// Without --from-lockfile, `carvel bake` regenerates the
+				// tile's own release from source (same as Step 1) and never
+				// touches Artifactory — it wouldn't exercise the lockfile/CI
+				// path this step is meant to test.
```
rizwanreza: Can we remove this and a bunch of comments in this PR? Don't think we need this many comments.

jobs/smoke_tests.tgz
rizwanreza, line 1: Don't think this file is being used anywhere?

internal/commands/carvel_bake.go
```diff
@@ -54,36 +58,68 @@ func (c CarvelBake) Execute(args []string) error {
...
+	} else if lockfilePresent {
+		return fmt.Errorf("failed to load Kilnfiles: %w", loadErr)
+	} else if c.Options.FromLockfile {
+		return fmt.Errorf("failed to load Kilnfiles (required for --from-lockfile): %w", loadErr)
+	} else {
+		// lockfilePresent is false here, so the lockfile is confirmed absent
+		// — that's what actually blocks additional_releases/--from-lockfile,
+		// regardless of whether a Kilnfile exists on its own (LoadKilnfiles
+		// always requires both, so we can't tell from loadErr alone).
+		c.outLogger.Printf("No Kilnfile.lock found — proceeding without additional_releases support")
```
rizwanreza: When `Kilnfile.lock` is absent, `loadErr` from `LoadKilnfiles` is silently swallowed here even if the *Kilnfile itself* (not the lock) is malformed — the comment above even says as much ("we can't tell from loadErr alone"). So a broken `Kilnfile` (bad `variable(...)` ref, syntax error, etc.) with no `Kilnfile.lock` yet just prints "No Kilnfile.lock found — proceeding without additional_releases support" instead of surfacing the real parse error.

`carvel_upload.go` already solves this correctly (calls `loadKilnfileOnly` in isolation first to tell which file actually failed) — worth reusing that pattern here, and in `carvel_rebake.go` which has the same gap around its own `loadErr` handling, rather than assuming "lockfile absent ⇒ any load error is fine to ignore."

No test currently covers "malformed Kilnfile + absent lockfile."

internal/commands/carvel_bake.go
```diff
@@ -19,9 +20,12 @@ type CarvelBake struct {
...
 type CarvelBakeOptions struct {
 	flags.Standard
-	SourceDirectory string `short:"s" long:"source-directory" description:"path to the Carvel tile source directory (defaults to current directory)"`
-	OutputFile      string `short:"o" long:"output-file"      description:"path to where the tile will be output" required:"true"`
-	Verbose         bool   `short:"v" long:"verbose"           description:"enable verbose output"`
+	SourceDirectory   string `short:"s"   long:"source-directory"   description:"path to the Carvel tile source directory (defaults to current directory)"`
+	OutputFile        string `short:"o"   long:"output-file"        description:"path to where the tile will be output" required:"true"`
+	Verbose           bool   `short:"v"   long:"verbose"            description:"enable verbose output"`
+	ReleasesDirectory string `short:"rd"  long:"releases-directory" default:"releases" description:"path to a directory containing/receiving additional_releases tarballs"`
+	SkipFetchReleases bool   `short:"sfr" long:"skip-fetch"         description:"skip fetching additional_releases; expect them already present in --releases-directory"`
```
rizwanreza, line 27: Minor UX inconsistency: `--releases-directory`/`--skip-fetch` get short aliases here (`-rd`, `-sfr`), but the identical flags on `carvel_publish.go` (`ReleasesDirectory`/`SkipFetch`) and `carvel_upload.go` have no short alias at all. Same concept, different typing cost depending on which carvel subcommand you're in. Worth either adding the shorts to publish/upload too, or dropping them here, so the three commands feel consistent.

(`-rd` itself is good — it matches classic `kiln bake`'s own `--releases-directory` short flag in `bake.go`, so that part is intentional reuse, not a new convention.)

internal/carvel/baker.go
```diff
@@ -855,6 +1006,98 @@ func (b *baker) generateJobFiles() error {
...
+	}
+	switch source.Configuration().Type {
+	case cargo.BOSHReleaseTarballSourceTypeS3, cargo.BOSHReleaseTarballSourceTypeArtifactory:
+		return filepath.Base(lockEntry.RemotePath)
+	default:
+		return defaultName
+	}
+}
+
+func findReleaseLock(lock cargo.KilnfileLock, name string) (cargo.BOSHReleaseTarballLock, bool) {
```
rizwanreza: Nit: this duplicates `cargo.KilnfileLock.FindBOSHReleaseWithName`, which this same PR already uses everywhere else (`carvel_bake.go`, `carvel_publish.go`, `carvel_rebake.go`, `carvel_helpers.go`). Could just call that and check the error instead of keeping a second lookup helper with a bool-return signature.

rizwanreza (commented):
Overall: solid, well-scoped extension of `carvel bake` — verified locally that it builds clean (`go build ./...`) and all unit tests pass (`go test ./internal/carvel/... ./internal/commands/... ./pkg/cargo/...`). The name-based lockfile lookups, the s3/artifactory filename-mismatch fix, and the explicit `--from-lockfile` opt-in are good, well-tested bug fixes.

Left a few inline notes:
- A malformed `Kilnfile` (as opposed to a malformed/missing `Kilnfile.lock`, which this PR already handles well) can still be silently swallowed in `carvel_bake.go`/`carvel_rebake.go` when no lockfile is present — `carvel_upload.go` already has the right pattern for disambiguating this.
- Flag-ergonomics nit: `-rd`/`-sfr` short aliases exist on `bake` but not on the identical `publish`/`upload` flags.
- Minor duplication: `findReleaseLock` in `baker.go` duplicates `cargo.KilnfileLock.FindBOSHReleaseWithName`.

None of these block the core functionality — the default (no-flags-beyond-required) path works correctly for tiles without `additional_releases`/hooks, and all three new flags degrade safely to prior behavior.

internal/commands/carvel_bake.go
```diff
@@ -54,36 +58,68 @@ func (c CarvelBake) Execute(args []string) error {
...
+		return fmt.Errorf("failed to load Kilnfiles (required for --from-lockfile): %w", loadErr)
+	} else {
+		// lockfilePresent is false here, so the lockfile is confirmed absent
+		// — that's what actually blocks additional_releases/--from-lockfile,
+		// regardless of whether a Kilnfile exists on its own (LoadKilnfiles
+		// always requires both, so we can't tell from loadErr alone).
+		c.outLogger.Printf("No Kilnfile.lock found — proceeding without additional_releases support")
+	}
 
+	if c.Options.FromLockfile {
```
notrepo05, line 90: I'm wondering if we could maintain backwards compatibility for existing users by defaulting to using the `Kilnfile.lock` pin for the generated (`baker.GetBoshReleaseName()`) bosh release, that way the release isn't regenerated when it's already pinned / the cache is used. I'm concerned about the impact to bake/publish on existing pipelines who use this command.

jobs/smoke_tests.tgz
ichandrabhatta: Will remove this file.

internal/commands/carvel_bake.go
ichandrabhatta: Made the needed changes.
Now, if kiln carvel bake is run without the explicit --from-lockfile flag, it will still check if the Kilnfile.lock is present and if it contains a pin for the tile's own generated BOSH release. If the pin is found, it will default to using the cached release from the lockfile instead of regenerating it from source. This ensures that existing pipelines relying on the cache are not impacted, addressing the PR comment.

internal/acceptance/carvel/carvel_workflow_test.go
ichandrabhatta: Cleaning up most of the comments except for a very few relevant ones.

internal/commands/carvel_bake.go
ichandrabhatta: Addressed this issue

internal/commands/carvel_bake.go
ichandrabhatta: fixed this mismatch.

internal/carvel/baker.go
ichandrabhatta: Fixed this functionality duplication.

rizwanreza (approved):

rizwanreza:
I would squash the commits, especially the ones that add and remove unneeded files since they don't need to be part of history.

internal/commands/carvel_helpers_test.go
```diff
@@ -0,0 +1,74 @@
...
+	initialLock := cargo.KilnfileLock{
+		Releases: []cargo.BOSHReleaseTarballLock{
+			{Name: "cf-cli", Version: "1.0.0"},
+		},
+	}
+	data, err := yaml.Marshal(&initialLock)
+	require.NoError(t, err)
+	require.NoError(t, os.WriteFile(lockfilePath, data, 0644))
+	require.NoError(t, os.Chmod(lockfilePath, 0000))
+	defer func() { _ = os.Chmod(lockfilePath, 0644) }()
```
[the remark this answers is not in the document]
ichandrabhatta: Non issue for now

internal/carvel/baker_test.go
```diff
@@ -705,6 +1090,39 @@ consumes:
 				Expect(err.Error()).To(ContainSubstring("missing required field 'name'"))
 			})
```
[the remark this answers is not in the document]
ichandrabhatta: This is fixed as a part of the updates done to the tests to address other comments

internal/commands/carvel_bake.go
ichandrabhatta: Addressed this issue.

notrepo05 (approved):

ram-pivot:
Since this PR was from a Forked repo the Github secrets for the workflows are not available. 

Cherry picked the commit from the forked repo/

Create Draft PR to run the PR checks - https://github.com/pivotal-cf/kiln/pull/675

ram-pivot:
PR Checks passed now:
https://github.com/pivotal-cf/kiln/actions/runs/31015789613/job/92339358948?pr=675

PR #676, opened 2026-08-18
Add kapp ordering annotations to carvel-baked k8s resources

fjagugar:
Carvel bake generates a bosh release whose registry-data job installs a ServiceAccount, ClusterRole, ClusterRoleBinding, Secret, and PackageInstall via an ERB template. kapp applied these with no ordering guarantee, so the PackageInstall could be created before its RBAC/values Secret existed, or deleted before the PackageInstall was torn down, leaving orphaned or permission-less resources.

Add kapp.k14s.io change-group/change-rule annotations so kapp installs the ServiceAccount, ClusterRole, ClusterRoleBinding, and Secret before the PackageInstall, and deletes them only after the PackageInstall is removed:

- ServiceAccount, ClusterRole, ClusterRoleBinding: change-group "kiln.tanzu.vmware.com/rbac"
- Secret: change-group "kiln.tanzu.vmware.com/secrets"
- PackageInstall: change-group "kiln.tanzu.vmware.com/packageinstall"

Each of the RBAC/Secret resources declares an installation rule to upsert before the PackageInstall and a cleanup rule to delete after it.

Also add tests covering the annotations on each generated resource in generateManifestTemplate.


internal/carvel/baker_test.go
```diff
@@ -44,6 +44,17 @@ func fileChecksum(path string) string {
 	return hex.EncodeToString(h.Sum(nil))
 }
 
+// templateDocument returns the YAML document within a multi-document
+// template whose "kind:" line matches the given kind exactly.
+func templateDocument(template, kind string) string {
+	for doc := range strings.SplitSeq(template, "---\n") {
+		if strings.Contains(doc, "kind: "+kind+"\n") {
```
drich10: 🤖 AI Review
---
🟢 **Suggestion**

`templateDocument` is subtly fragile for `"ClusterRole"`: `strings.Contains(doc, "kind: ClusterRole\n")` also matches inside the ClusterRoleBinding document because the `roleRef:` block contains `kind: ClusterRole` on an indented line — `Contains` is a substring match so the indent doesn't prevent it.

It currently works only because the ClusterRole document appears *before* ClusterRoleBinding in the template, so `SplitSeq` hits it first. If the document order ever changes, or a new resource references `ClusterRole`, this helper silently returns the wrong document.

A safer match would anchor to the start of the line:
```go
if strings.Contains(doc, "\nkind: "+kind+"\n") || strings.HasPrefix(doc, "kind: "+kind+"\n") {
```

Not blocking — the test is correct today — but worth hardening.

ram-pivot:
Since PR checks don't run on the forked repo, created test PR with the changes  - https://github.com/pivotal-cf/kiln/pull/684/

ram-pivot:
**Automated review — this comment was generated by a bot (LLM-based review), not a human reviewer. Please weigh it accordingly.**

Reviewed the diff (`internal/carvel/baker.go`, `internal/carvel/baker_test.go`). The kapp change-rule grammar used (`<upsert|delete> <before|after> <upserting|deleting> <group>`, with `.installation`/`.cleanup` as free-form change-rule suffixes) matches kapp's documented annotation format, and the direction of each rule matches the stated intent (RBAC/Secret up before PackageInstall, down only after). `go test ./internal/carvel/...` passes, and I didn't find any golden-fixture/checksum test that this diff would break.

Two things worth a look before merge:

1. **Change-group names aren't scoped per `package_installs` entry** (`baker.go` lines 667/676/689/707/731). `generateManifestTemplate` hardcodes the literal groups `kiln.tanzu.vmware.com/rbac`, `.../secrets`, and `.../packageinstall` regardless of `entry`. `metadata.PackageInstalls` is a `[]string`, so a tile with two or more package installs (A and B) will have *all* entries' ServiceAccounts/ClusterRoles/ClusterRoleBindings/Secrets collapse into the same three groups. kapp's rule engine orders a resource against every group member present in the current change-set — not just the members belonging to the same entry. So a simultaneous multi-entry teardown couples unrelated entries: if B's PackageInstall gets stuck deleting (e.g. a kapp-controller finalizer hang), A's otherwise-unrelated RBAC/Secret cleanup is blocked behind it too, even though A has nothing to do with B. Per-entry install/delete ordering for a single entry still works correctly (each entry's own PackageInstall satisfies its own RBAC's rule), so this isn't a regression of the bug described in the PR body — it's a coupling side effect that only shows up with 2+ entries, which isn't covered by any current test. Might be worth namespacing the groups by entry, e.g. `kiln.tanzu.vmware.com/rbac-<entry>`.

2. **Minor duplication**: the 3-line annotation block (`change-group` + `change-rule.installation` + `change-rule.cleanup`) is repeated as a literal string four times (three of them byte-for-byte identical, for the `rbac` group). Extracting a small helper (e.g. `kappOrderingAnnotations(group, targetGroup string) string`) would remove the risk of the copies drifting on a future edit — right now only the new tests would catch a slipped copy.

Nothing else stood out; the annotation placement/indentation in the generated YAML is correct, and the new `templateDocument` test helper's `kind: X\n` matching correctly avoids `ClusterRole` matching `ClusterRoleBinding`.

ram-pivot:
Test PR passed the checks - https://github.com/pivotal-cf/kiln/pull/684

ram-pivot:
Another test PR with updated changes - https://github.com/pivotal-cf/kiln/pull/685

ram-pivot:
https://github.com/pivotal-cf/kiln/pull/685 PR checks passed.

ram-pivot:
I have not tested the changes with relevant environment but the changes look good to me. 

Hope @fjagugar already validated the ordering changes.

PR #679, opened 2026-08-19
fix(carvel): allow carvel bake without Kilnfile

### LINKED ISSUE #677: Latest kiln version requires Kilnfile to be present when running kiln carvel bake
### Describe the bug

In previous version 0.117.0 I was able to run `kiln carvel bake` command without having a Kilnfile present in the folder where my tile code was present.

However, while doing the same thing with the current kiln code present in main branch I get an error:

```
❯ /tmp/kiln carvel bake --source-directory /tmp/tile --output-file /tmp/output-tile/output.pivotal
2026/08/18 09:51:51 could not execute "bake": failed to load Kilnfile: failed to open Kilnfile: open /tmp/tile/Kilnfile: no such file or directory
```

After placing an empty Kilnfile in the tile source folder it worked.

### Reproduction steps

1. Place your tile code in a folder
2. Do not include / remove Kilnfile from that folder
3. Run kiln carvel bake using latest kiln version from main branch
4. Command fails complaining about missing Kilnfile

### Expected behavior

A new pivotal file is generated without issues

### Additional context

_No response_

ichandrabhatta:
Fixes a regression where `kiln carvel bake` would fail if a Kilnfile was not present in the tile directory. This restores backward compatibility for tiles that do not use `additional_releases`.

Also refines the error messages for the `--from-lockfile` flag and adds an acceptance test to lock in the expected behavior.

Fixes #677
JIRA-ID: TNZ-132681

### DIFF
--- a/internal/acceptance/carvel/carvel_bake_test.go
+++ b/internal/acceptance/carvel/carvel_bake_test.go
@@ -154,6 +154,22 @@ var _ = Describe("carvel bake command", func() {
 		Eventually(verifySession.Out).Should(gbytes.Say("No errors detected"))
 	})
 
+	Context("when no Kilnfile is present", func() {
+		It("bakes successfully without additional_releases support", func() {
+			err := os.Remove(filepath.Join(inputPath, "Kilnfile"))
+			Expect(err).NotTo(HaveOccurred())
+
+			command := exec.Command(pathToMain, commandWithArgs...)
+
+			session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(session, "60s").Should(gexec.Exit(0))
+			Eventually(session.Out).Should(gbytes.Say("No Kilnfile found — proceeding without additional_releases support"))
+			Eventually(session.Out).Should(gbytes.Say("Done! Baked"))
+		})
+	})
+
 	Context("failure cases", func() {
 		Context("when the output-file flag is not provided", func() {
 			It("prints an error and exits 1", func() {
@@ -197,7 +213,7 @@ var _ = Describe("carvel bake command", func() {
 				Expect(err).NotTo(HaveOccurred())
 
 				Eventually(session).Should(gexec.Exit(1))
-				Eventually(session.Err).Should(gbytes.Say("Kilnfile"))
+				Eventually(session.Err).Should(gbytes.Say("base.yml"))
 			})
 		})
 	})
--- a/internal/commands/carvel_bake.go
+++ b/internal/commands/carvel_bake.go
@@ -46,6 +46,10 @@ func (c CarvelBake) Execute(args []string) error {
 		return err
 	}
 
+	if _, err := os.Stat(sourcePath); err != nil {
+		return fmt.Errorf("source directory not found: %s", sourcePath)
+	}
+
 	targetPath, err := filepath.Abs(c.Options.OutputFile)
 	if err != nil {
 		return fmt.Errorf("failed to resolve output file path: %w", err)
@@ -66,11 +70,14 @@ func (c CarvelBake) Execute(args []string) error {
 	_, lockfileStatErr := os.Stat(c.Options.KilnfileLockPath())
 	lockfilePresent := lockfileStatErr == nil
 
+	_, kilnfileStatErr := os.Stat(kilnfilePath)
+	kilnfilePresent := kilnfileStatErr == nil
+
 	kf, kl, loadErr := c.Options.LoadKilnfiles(nil, nil)
 	if loadErr == nil {
 		kilnfile = kf
 		kilnfileLock = kl
-	} else {
+	} else if kilnfilePresent {
 		kfOnly, kfErr := loadKilnfileOnly(c.Options.Standard)
 		if kfErr != nil {
 			return fmt.Errorf("failed to load Kilnfile: %w", kfErr)
@@ -80,11 +87,19 @@ func (c CarvelBake) Execute(args []string) error {
 		if lockfilePresent {
 			return fmt.Errorf("failed to load Kilnfile.lock: %w", loadErr)
 		} else if c.Options.FromLockfile {
-			return fmt.Errorf("failed to load Kilnfiles (required for --from-lockfile): %w", loadErr)
+			return fmt.Errorf("failed to load Kilnfile.lock (required for --from-lockfile): %w", loadErr)
 		} else {
 			c.outLogger.Printf("No Kilnfile.lock found — proceeding without additional_releases support")
 		}
-	}
+		} else {
+			if c.Options.FromLockfile {
+				return fmt.Errorf("kilnfile not found at %s (required for --from-lockfile)", kilnfilePath)
+			}
+			if lockfilePresent {
+				return fmt.Errorf("kilnfile not found at %s but Kilnfile.lock is present; refusing to bake from source and silently ignore the pinned lock", kilnfilePath)
+			}
+			c.outLogger.Printf("No Kilnfile found — proceeding without additional_releases support")
+		}
 
 	var useLockfile bool
 	if c.Options.FromLockfile {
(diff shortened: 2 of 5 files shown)


internal/commands/carvel_bake.go
```diff
@@ -73,16 +74,24 @@ func (c CarvelBake) Execute(args []string) error {
...
 		if kfErr != nil {
-			return fmt.Errorf("failed to load Kilnfile: %w", kfErr)
-		}
-		kilnfile = kfOnly
-
-		if lockfilePresent {
-			return fmt.Errorf("failed to load Kilnfile.lock: %w", loadErr)
-		} else if c.Options.FromLockfile {
-			return fmt.Errorf("failed to load Kilnfiles (required for --from-lockfile): %w", loadErr)
+			if errors.Is(kfErr, os.ErrNotExist) {
```
ram-pivot: `errors.Is(kfErr, os.ErrNotExist)` here conflates a missing Kilnfile with any missing file the loader touches — e.g. a typo'd `--variables-file`. If the Kilnfile is valid but `--variables-file` points at a moved/typo'd path, `FromPathsAndPairs` -> `os.Open` fails, gets wrapped with `%w`, and still satisfies `errors.Is(err, os.ErrNotExist)`. This logs "No Kilnfile found — proceeding without additional_releases support" and continues with a zero-value Kilnfile instead of surfacing the real problem — bake succeeds silently but is missing `additional_releases` the user expected.

internal/commands/carvel_rebake.go
```diff
@@ -98,9 +99,12 @@ func (c CarvelReBake) Execute(args []string) error {
 	} else {
 		kfOnly, kfErr := loadKilnfileOnly(c.Options.Standard)
 		if kfErr != nil {
-			return fmt.Errorf("failed to load Kilnfile: %w", kfErr)
+			if !errors.Is(kfErr, os.ErrNotExist) {
```
ram-pivot: Same `ErrNotExist` conflation as `carvel_bake.go:77`, but worse here: there's no log message at all when this silently drops the Kilnfile, so a missing/typo'd `--variables-file` fails silently with zero diagnostic output.

internal/commands/carvel_bake.go
```diff
@@ -73,16 +74,24 @@ func (c CarvelBake) Execute(args []string) error {
...
+				}
+				c.outLogger.Printf("No Kilnfile found — proceeding without additional_releases support")
+			} else {
+				return fmt.Errorf("failed to load Kilnfile: %w", kfErr)
+			}
 		} else {
-			c.outLogger.Printf("No Kilnfile.lock found — proceeding without additional_releases support")
+			kilnfile = kfOnly
+
+			if lockfilePresent {
```
ram-pivot: `bake` and `rebake` now disagree on identical state. Delete `Kilnfile` but leave `Kilnfile.lock`: in `carvel_bake.go`, `kfErr` is `ErrNotExist` so it logs and continues; `lockfilePresent && loadErr == nil` is false (since `loadErr != nil`), so it silently bakes from source, discarding the pinned lock with no warning. In `carvel_rebake.go` (line ~113), the identical state hard-fails with "failed to load Kilnfile.lock". Pre-PR both hard-failed; now the two commands diverge on the same input.

internal/commands/carvel_bake.go
```diff
@@ -73,16 +74,24 @@ func (c CarvelBake) Execute(args []string) error {
...
-
-		if lockfilePresent {
-			return fmt.Errorf("failed to load Kilnfile.lock: %w", loadErr)
-		} else if c.Options.FromLockfile {
-			return fmt.Errorf("failed to load Kilnfiles (required for --from-lockfile): %w", loadErr)
+			if errors.Is(kfErr, os.ErrNotExist) {
+				if c.Options.FromLockfile {
+					return fmt.Errorf("failed to load Kilnfile (required for --from-lockfile): %w", kfErr)
+				}
+				c.outLogger.Printf("No Kilnfile found — proceeding without additional_releases support")
```
ram-pivot: A nonexistent/mistyped `--source-directory` also gets misdiagnosed as "no Kilnfile" here and fails later with an unrelated, confusing error (deep inside `baker.Bake`/`ParseMetadata`) instead of failing fast with the previous clear "failed to load Kilnfile: ... no such file or directory". Worth noting the existing acceptance test for this only asserts `exit(1)`, not message content, so this regression wouldn't be caught.

internal/acceptance/carvel/carvel_bake_test.go
```diff
@@ -154,6 +154,23 @@ var _ = Describe("carvel bake command", func() {
 		Eventually(verifySession.Out).Should(gbytes.Say("No errors detected"))
 	})
 
+	Context("when no Kilnfile is present", func() {
```
ram-pivot, line 157: Coverage gap: only the "Kilnfile literally absent, all variable pairs valid" happy path is tested. Nothing exercises the non-`ErrNotExist` branch (should still hard-fail) or the "`ErrNotExist` but caused by something other than a missing Kilnfile" case from the comments above — the actual bug scenario has no test that would catch it.

ram-pivot (commented):
Reviewed by Ramkumar Vengadakrishnan — automated (bot) review generated by Claude Code.

Root issue: `errors.Is(kfErr, os.ErrNotExist)` is checked against an aggregate multi-step loader error in `carvel_bake.go` / `carvel_rebake.go`. That's too broad — it conflates "Kilnfile itself is missing" with any other ENOENT the loader can hit (a typo'd `--variables-file`, a bad `--source-directory`), so unrelated errors get silently swallowed as "no Kilnfile." It also makes `bake` and `rebake` diverge on identical on-disk state (Kilnfile absent, Kilnfile.lock present): `bake` silently ignores the lock and builds from source, `rebake` hard-fails.

Suggested fix: reuse the direct `os.Stat(kilnfilePath)` check already present in `carvel_upload.go:57` instead of inferring "Kilnfile missing" from `errors.Is` on the loader's error.

Also: the new acceptance test intended to lock this in doesn't currently run in CI — no workflow installs the `bosh` CLI, so `BeforeSuite` in the carvel acceptance suite always `Skip()`s on the runners.

Inline comments below for specifics.

ichandrabhatta:
Thanks for the detailed review @ram-pivot — all points were valid. Addressed in the latest commit.

**Root fix:** replaced `errors.Is(kfErr, os.ErrNotExist)` with a direct `os.Stat(kilnfilePath)` check in both `carvel bake` and `carvel rebake` — the same pattern already used in `carvel_upload.go`. The "is the Kilnfile missing?" decision no longer piggybacks on the loader's aggregate error, so a missing/typo'd `--variables-file` (or bad `--source-directory`) can't be misdiagnosed as "no Kilnfile."

Per-comment:

- **`carvel_bake.go` — ErrNotExist conflates missing Kilnfile with missing `--variables-file`:** fixed. A present-but-unloadable Kilnfile (parse error, bad variables file, unreadable lock) now surfaces its real error instead of silently baking without `additional_releases`.
- **`carvel_rebake.go` — same conflation:** fixed with the same `os.Stat` gate.
- **bake/rebake divergence on "Kilnfile absent + lock present":** resolved by making both **hard-fail** on that state (bake now errors "Kilnfile not found … but Kilnfile.lock is present; refusing to bake from source and silently ignore the pinned lock"; rebake already hard-failed via its lock block).
- **Bad `--source-directory` misdiagnosed:** added an up-front `os.Stat(sourcePath)` check that fails fast with "source directory not found".
- **Coverage gap:** added three bosh-free command-level tests (bad `--variables-file`, absent-Kilnfile-with-lock, nonexistent `--source-directory`) plus an analogous rebake test. These return before `baker.Bake`, so unlike the acceptance suite they run in CI without the `bosh` CLI.

`go build`/`go vet` clean; `go test ./internal/commands/... ./internal/acceptance/carvel/...` passes (151 specs, 0 failed).

Note: the acceptance suite's `BeforeSuite` still `Skip()`s when `bosh` isn't on PATH, so the full end-to-end bake test doesn't run in CI. The new command-level tests cover the regression logic in CI; installing `bosh` on the runners to also exercise the acceptance path can be a follow-up if desired.

ram-pivot:
Created Test PR - https://github.com/pivotal-cf/kiln/pull/681 with commits from this PR to run the checks.

ram-pivot:
Created test PR with changes - https://github.com/pivotal-cf/kiln/pull/683/ to run the PR checks. 

PR Checks passed on that PR. Merging this PR.

ram-pivot:
Javier confirmed the changes work by testing. So merging the PR.

PR #682, opened 2026-08-24
TNZ-132747: Make carvel bake and upload reproducible for BOSH releases

ram-pivot:
`bosh create-release` stamps real wall-clock timestamps into the gzip header of every nested jobs/*.tgz and packages/*.tgz blob (and those blobs embed real filesystem mtimes), making the blob sha256 - and therefore release.MF and the outer release tarball - non-deterministic across otherwise-identical bakes.

Add canonicalizeBoshRelease, called from baker.createBoshRelease, which normalizes mtime/uid/gid/ownership/PAX-xattrs/format on the outer tarball and every nested blob, then patches the sha1 (sha256-prefixed) checksums in release.MF to match the re-packed blobs. This mirrors the fixed epoch already used for the outer .pivotal zip in internal/commands/bake.go.

ai-assisted=yes
JIRA-ID: TNZ-132747


internal/acceptance/carvel/carvel_bake_test.go
```diff
@@ -23,9 +25,6 @@ func TestCarvelAcceptance(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	if _, err := exec.LookPath("bosh"); err != nil {
```
notrepo05, line 26: Good catch

internal/carvel/canonical_release.go
```diff
@@ -0,0 +1,449 @@
...
+			if hdr.Mode&0o111 != 0 {
+				hdr.Mode = hdr.Mode&0o7000 | 0o755
+			} else {
+				hdr.Mode = hdr.Mode&0o7000 | 0o644
+			}
+		}
+
+		hdr.PAXRecords = nil
+		hdr.Xattrs = nil //nolint:staticcheck // deprecated; cleared alongside PAXRecords
+		hdr.Format = tar.FormatUnknown
```
drich10: 🤖 AI Review
---
🟡 **Should Fix**

`tar.FormatUnknown` lets Go auto-select between USTAR and GNU per-entry based on name length (USTAR requires Name ≤ 100 bytes; longer names force GNU extension headers). This can produce a **mixed-format tarball** — some entries encoded as USTAR blocks, others as GNU — and the selection logic is internal to the Go runtime so it could change across Go versions.

Since the entire PR is about byte-for-byte reproducibility, pinning to an explicit format is the right call here:

```go
hdr.Format = tar.FormatGNU
```

GNU format handles long names natively, is well-understood by all BOSH tooling, and eliminates any ambiguity about which format variant the tar writer will choose.

internal/carvel/canonical_release.go
```diff
@@ -0,0 +1,449 @@
...
+		_ = os.Remove(manifestEntry.path)
+		untrack(manifestEntry.path)
+		entries[manifestFileName] = tarEntry{header: manifestEntry.header, path: mf.Name(), size: int64(len(patched))}
+	}
+
+	out, err := os.CreateTemp(spoolDir, filepath.Base(tarballPath)+".canonical-*.tmp")
+	if err != nil {
+		return fmt.Errorf("failed to write canonical bosh release tarball: %w", err)
+	}
+	tmp := out.Name()
```
drich10: 🤖 AI Review
---
🟡 **Should Fix**

Every other spool file in this function is tracked in `spoolFiles` for deferred cleanup, but `tmp` (the final output file) is not. It relies on an explicit `_ = os.Remove(tmp)` in each error branch. If the code ever panics between `os.CreateTemp` and `os.Rename` — or a new error branch is added that forgets the remove — `tmp` leaks on disk in the tarball's directory.

Two clean alternatives:

1. Use `defer os.Remove(tmp)` immediately after the `CreateTemp` call. After a successful `os.Rename(tmp, tarballPath)` the path `tmp` no longer exists, so the deferred remove is a no-op (ignoring the error).

2. Follow the existing pattern: `track(tmp)` after `CreateTemp`, then `untrack(tmp)` immediately before `os.Rename`. This makes the invariant self-documenting.

Option 1 also lets you remove the four repeated `_ = os.Remove(tmp)` lines below.

internal/carvel/canonical_release.go
```diff
@@ -0,0 +1,449 @@
...
+	if err != nil {
+		return "", 0, "", err
+	}
+
+	return writeCanonicalBlob(entries, spoolDir)
+}
+
+type countingWriter struct{ n int64 }
+
+func (w *countingWriter) Write(p []byte) (int, error) {
```
drich10: 🤖 AI Review
---
🟢 **Suggestion**

`countingWriter.Write` discards all bytes — it's a counting-only sink used inside an `io.MultiWriter` alongside the real `dst` and `h` writers. A one-line comment prevents the next reader from wondering why nothing is stored:

```go
// countingWriter counts bytes written; it is always paired with a real writer
// in an io.MultiWriter, so it intentionally discards the data.
type countingWriter struct{ n int64 }
```

internal/carvel/canonical_release.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	if hasManifest {
+		patched, unpatched, err := patchReleaseManifestChecksums(manifestData, newSHA)
+		if err != nil {
+			return fmt.Errorf("failed to patch release.MF: %w", err)
+		}
+
+		if len(unpatched) > 0 {
+			return fmt.Errorf("repacked blob(s) %s have no sha1 field in release.MF to patch; "+
```
drich10: 🤖 AI Review
---
🟢 **Suggestion**

The `unpatched` / `remaining` keys are in the form `"jobs/foo"` (without `.tgz`), so the error message reads:

> repacked blob(s) jobs/foo have no sha1 field …

A user debugging a broken `release.MF` would find the actual filename (`jobs/foo.tgz`) more actionable. Consider appending `.tgz` when building the display list, or include both the key and the filename in the message.

internal/carvel/baker.go
```diff
@@ -1126,6 +1131,10 @@ func (b *baker) createBoshRelease() error {
 		return err
 	}
 
+	if err := canonicalizeBoshRelease(finalTarball); err != nil {
```
notrepo05, line 1152: There's a conceivable (though very unlikely) edge case when a user has an existing carvel-based tile and generated bosh release. If they rebuild a version (exact same bundle.tar etc), _and_ don't reuse an existing cached release in artifactory, then the bosh version will remain the same but the actual bosh release sha will change due to canonicalization. One solution is to just write a e.g. canonicalization version into the sha generation (example below), this should avoid this tiny possibility. Though this will cause all generated bosh releases to change upon upgrade.

internal/carvel/baker.go
notrepo05, line 1161:
```suggestion
    canonicalizationVersion := "1.0"
	h := sha256.New()
	// todo: error checking!
	_, _ = fmt.Printf(h, "canonicalization-version: %s", canonicalizationVersion)
```

notrepo05 (changes requested):
Thanks for the PR!  The coverage is excellent. I might have found a minor edge case during upgrades for pre-existing kiln carvel tiles when regenerating a bosh release. It's very unlikely, but I wanted to raise it for review. Thanks!

notrepo05 (approved):
LGTM. Based on actual usage patterns the comment I left might not apply, so leaving just for consideration. Thanks!

PR #689, opened 2026-09-17
Add replicable field support to carvel tile metadata

drich10:
## Summary

- Regular (BOSH) tiles can set `replicable: true` in `base.yml` and it passes through to the baked metadata generically via template interpolation
- Carvel/k8s tiles use a strongly-typed `Metadata` struct — the `replicable` field was silently dropped during bake since it wasn't defined
- Adds `Replicable bool \`yaml:"replicable,omitempty"\`` to `models.Metadata` and `models.MetadataOut`, and copies it in `generateBaseYaml()` to match parity with the regular tile bake path

## Test plan

- [ ] New Ginkgo specs in `internal/carvel/baker_test.go` cover both cases: field present → passes through; field absent → omitted from output
- [ ] `go test ./internal/carvel/... ./internal/commands/... ./pkg/...` — 314 passed, 0 failed
- [ ] Manually verify by adding `replicable: true` to a k8s tile `base.yml` and running `kiln carvel-bake`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/carvel/baker_test.go
```diff
@@ -574,6 +575,58 @@ product_version: "0.1.0"
 			Expect(err.Error()).To(ContainSubstring("missing required field 'type'"))
 		})
 	})
+
+	Context("generateBaseYaml", func() {
+		It("passes replicable: true from input metadata to output base.yml", func() {
```
drich10, line 600: 🤖 AI Review
---
🟢 **Suggestion**

These tests verify that `generateBaseYaml()` propagates `Replicable: true` from `models.Metadata` to `models.MetadataOut` and serializes it to `base.yml`.

Consider also adding a test for the YAML unmarshaling side (e.g. via `baker.ParseMetadata` or directly unmarshaling a YAML string with `replicable: true` into `models.Metadata`). This would ensure full coverage across both the ingestion from `base.yml` and emission to `base.yml`.

rizwanreza (approved):
LGTM!

PR #690, opened 2026-09-21
TNZ-153678: Anchor default --releases-directory to source dir, not CWD

ram-pivot:
carvel bake/upload/publish/rebake defaulted ReleasesDirectory via a jhanda `default:"releases"` tag, so an unset flag resolved relative to the process's CWD instead of --source-directory, and an explicit --releases-directory releases was silently treated the same as unset. fetchAdditionalReleases also never created ReleasesDirectory before downloading into it, masked in most specs by BeforeEach pre-creating the dir.

Replace the tag default with resolveReleasesDirectory(), which treats only "" as unset and anchors it to sourcePath/releases; any explicit value (including the literal "releases") now resolves like a normal override. Add the missing os.MkdirAll for ReleasesDirectory in fetchAdditionalReleases, with a regression test that deliberately does not pre-create the cache dir. Update carvel_bake/upload/publish flag descriptions and TILE_AUTHOR_GUIDE.md accordingly.

ai-assisted=yes


rizwanreza (approved):
Approving this to unblock. I think we should perhaps think about the design of this work post Europa.

notrepo05 (approved):

